Issue #19, opened 2019-10-09
Links broken


lilyminium:
[This link] (notation) doesn't work. Also, the relative links in the quick start guide should be updated.

PR #21, opened 2019-10-10
fixed links

lilyminium:
* fixed links in installation
* changed relative links in notebooks to absolute URLs


lilyminium:
Fixes #19

PR #33, opened 2019-12-13
updated section on MDAnalysisTests in Installation Guide

orbeckst:
- note that MDAnalysisTests is needed for many examples in this guide
- corrected pip upgrade


lilyminium (approved):
LGTM, thanks for the clarification and correction.

PR #34, opened 2019-12-28
RMS notebooks

lilyminium:
Jupyter notebooks for rms.rmsd, rms.RMSD, and rms.RMSF. 

I thought pairing notebook tutorials with short pages in the user guide would be a useful approach to the analysis packages, but it's not without flaws. Suggestions would be appreciated!

Pros:
* Users can download the notebooks and run it themselves
* *or* view it online
* Examples get checked (edit: although only when the notebook gets executed, not every time the documentation is built)

Cons
* diffing Jupyter notebooks is not convenient. A minor typo correction gives a totally different file. This could be hard to maintain down the line.
* duplication of content: I do think the individual classes should have minimal working examples in their documentation
* Jupyter notebooks are still quite limited re: footnotes, citations, and the 'alert' boxes (I think -- I haven't looked too deeply into this.)

My approach so far has been throwing everything related into one tutorial with multiple sections. Another approach is to have short notebooks for each scenario (which I believe was suggested in the original GSOD proposals)

I'm also not sure that the `nglview` visualisation is very useful, and it's an extra package. The online visualisations will not be able to play the trajectory, due to not having access to the file.

I would also like to know if I should keep up with the version notes (e.g. in the RMSF notebook, the differences in loading coordinates between 0.17.0 and 0.19.0)


lilyminium:
Viewable online:
 - [RMSD](https://lilyminium.github.io/UserGuide/examples/analysis/rmsd.html)
 - [RMSF](https://lilyminium.github.io/UserGuide/examples/analysis/rmsf_v0.17.0_to_v0.20.1.html)

kain88-de:
have you tried [nbdime](https://nbdime.readthedocs.io/en/latest/) to view diffs of notebooks?

lilyminium:
Thanks @kain88-de — nbdime looks very helpful for diffing locally, but as far as I can tell there’s no way to integrate it into GitHub for easy viewing of the changes made in pull requests, which would be the main concern

hainm:
> The online visualisations will not be able to play the trajectory, due to not having access to the file.

@lilyminium Just FYI, you can store some frames' coordinates in embed html file: https://github.com/arose/nglview/pull/687

lilyminium:
Making this part of a more cohesive Analysis pull request.

PR #38, opened 2020-02-07
[WIP] Analysis

lilyminium:
Closes #4, #7, most of #29

This is taking much longer than I thought it would. I would appreciate any feedback on what I have so far, what I should prioritise next, and also any changes I should make before I try to finish the rest of the notebooks.

Every notebook aims to show a useful use case of a class or function in MDAnalysis, point out variables that users will likely need to change, visualise the results, and interpret them in some way.

**Notebooks on Analysis modules**
- base
    - [x] using [AnalysisFromFunction, analysis_class, AnalysisBase to write your own](https://lilyminium.github.io/UserGuide/examples/analysis/custom_trajectory_analysis.html)
- [align](https://lilyminium.github.io/UserGuide/examples/analysis/alignment_and_rms/README.html)
    - [x] alignto
    - [x] AlignTraj
        - [x] mention `match_atoms` (new in 0.21.0 or 1.0)
    - [x] Aligning a trajectory to various frames
    - [x] AverageStructure (new in 0.21.0 or 1.0)
- [rms](https://lilyminium.github.io/UserGuide/examples/analysis/alignment_and_rms/README.html)
    - [x] rmsd
    - [x] RMSD
    - [x] RMSF
    - [x] pairwise RMSD
- [contacts](https://lilyminium.github.io/UserGuide/examples/analysis/distances_and_contacts/README.html)
    - [x] Contacts (1D) (notebook currently for develop version)
    - [x] q1q2 (2D)
    - [x] Writing a custom analysis
- [distances](https://lilyminium.github.io/UserGuide/examples/analysis/distances_and_contacts/README.html)
    - [x] dist
    - [x] distance_array
    - [x] self_distance_array
    - [ ] between -- this is such a small function... might put it as a note in the Selection Language page
- [ ] helanal is very broken. Low priority
- psa
    - [x] [PSAnalysis](https://lilyminium.github.io/UserGuide/examples/analysis/trajectory_similarity/psa.html)
    - ~[ ] PSAPair~ don't think people will use this much 
    - ~[ ] Pair~ don't think people will use this much
- [encore](https://lilyminium.github.io/UserGuide/examples/analysis/trajectory_similarity/README.html)
    - [x] hes (harmonic ensemble similarity) (currently broken, waiting on https://github.com/MDAnalysis/mdanalysis/pull/2484)
    - [x] ces (clustering ensemble similarity)
    - [x] dres (dimension reduction ensemble similarity)
    - [x] ces_convergence
    - [x] dres_convergence
[...]

PR #52, opened 2020-02-29
Analysis notebooks for 0.20.1

lilyminium:
Peeled off some analysis notebooks from #38 and [edited them for version 0.20.1](https://lilyminium.github.io/mda-userguide-dev/examples/README.html).

Having this available seems better than passing around my fork of the user guide that has tutorials for code that isn't released or merged yet.


orbeckst:
This is a good idea. I haven't looked at the actual notebooks, though. Presumably, you will replace them with the 1.0 content?

(I am a bit short on time right now so I am not sure that I will be able to properly review 20+ notebooks in a timely manner... unless there are objections, I would say go ahead and fix issues as they come up.)

lilyminium:
Thanks @orbeckst, I'll merge later today unless there are objections. Yes, #38 still contains the original 1.0.0-targeted notebooks so they can be replaced when the relevant code is released. Should we consider having a ``develop`` version of the user guide similar to the ``develop`` docs?

Issue #63, opened 2020-04-06
instance of pip installed outside the current conda environment is used.


ss62171:
**Actual Behavior** 
According to [User guide](https://www.mdanalysis.org/UserGuide/contributing_code.html#building-mdanalysis) if we install MDA using `conda` environment and say I try to install some dependency using `pip` without installing it in my `conda` environment and it might be possible to have `pip` installed outside a `conda` environment which causes these module to install in global env.

**Expected Behavior**
using `pip` these packages must install in `conda` environment.

**Solution**
```python
# if using pip
pip install cython numpy mmtf-python mock six biopython \
 networkx cython matplotlib scipy griddataformats \
 hypothesis gsd codecov "seaborn>=0.7.0,<=0.9" \
 netcdf4 scikit-learn "joblib>=0.12" parmed psutil pytest
```
add a note at top of this like:
if using pip make sure you have `pip` installed in your `conda` environment

RMeli:
My understanding of `# if using pip` is that it refers to `Creating a development environment - With pip and virtualenv`, so in that case `conda` would not be installed in the first place. In the same cell you have the same instructions for `conda` (why use `pip` if you are within a `conda` environment and if packages can be installed with `conda`?):

```bash
# if using conda
conda install -c biobuilds -c conda-forge \
 cython numpy mmtf-python mock six biopython \
 networkx cython matplotlib scipy griddataformats \
 hypothesis gsd codecov "seaborn>=0.7.0,<=0.9" \
 clustalw=2.1 netcdf4 scikit-learn "joblib>=0.12"\
 psutil pytest
# if using conda with python 3.7 or 3.8, also run
conda install -c conda-forge parmed
``` 

However, I do agree that `pip` is needed to install MDanalysis even when using `conda` (and some packages are only present on PyPI), therefore adding `pip` to the list of packages installed with `conda` might be useful.

RMeli:
This should be fixed with #64 .

PR #64, opened 2020-04-08
More details on dependencies

### LINKED ISSUE #62: Add dependencies to installation
Edit: from https://github.com/MDAnalysis/mdanalysis/issues/2644

**Describe the bug**
I think you need cython and numpy at minimum to build from source, but the installation instructions [do not mention either](https://www.mdanalysis.org/UserGuide/installation.html#development-versions).

**To fix**
Add dependencies, and possibly a link to the contributing docs.

RMeli:
Fix #62 and #63

### DIFF
--- a/doc/source/contributing_code.rst
+++ b/doc/source/contributing_code.rst
@@ -209,21 +209,25 @@ and activated your virtual environment. First we need to install dependencies:
 
         # if using conda
         conda install -c biobuilds -c conda-forge \
-         cython numpy mmtf-python mock six biopython \
-         networkx cython matplotlib scipy griddataformats \
-         hypothesis gsd codecov "seaborn>=0.7.0,<=0.9" \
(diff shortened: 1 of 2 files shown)


lilyminium:
Thanks @RMeli. I was under the impression that `conda` [already installed pip](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-pkgs.html#installing-non-conda-packages)? I know each of my conda environments has its own `pip` binary, although I use Anaconda instead of Miniconda. 

```bash
$ which pip
/Users/lily/anaconda3/envs/mda/bin/pip
```

RMeli:
I just tested it quickly and Miniconda also seems to install `pip` when creating a new environment. However,  I definitely had this mismatch between `pip` and the `conda` environment before so it definitely doesn't hurt to install it explicitly. =)

Issue #67, opened 2020-06-07
Dev docs build


richardjgowers:
I was going to poke one of the Gsoc students to write docs on their feature, but the feature won’t ship for a while. Am I right that the user guide is published instantly? And how much of a hassle would it be to have a dev/master split like we have on the main repo, with dev pushing to a different location? I think @orbeckst had thoughts along these lines too...

@lilyminium probably know something here

lilyminium:
Yeah, @orbeckst and I talked about this briefly in [another issue](https://github.com/MDAnalysis/UserGuide/issues/66#issuecomment-638886192). When I was doing GSoD I kept (and still have) a little in-development version [on my own fork](https://lilyminium.github.io/UserGuide/) but we should have an official "dev" version of our own. 

@orbeckst suggested names like UserGuideDev, DevUserGuide, LatestUserGuide; it might be neater if you just have a mirrored subdomain for dev.mdanalysis.org/docs and dev.mdanalysis/userguide.

lilyminium:
He also brought up RTD versioning which I was going to look into but have not yet -- this would be really useful for people who need a particular version of MDA.

richardjgowers:
Hm, we're using a travis deploy action, but then we can only associate one site with one repo, so not sure if we can have 2 different builds from one repo.

lilyminium:
I checked out RTD versioning. Not sure how releases work just yet -- although `latest` is pegged to `master`, I think -- but you can build different docs for different branches. I've got one up for an "rtd" branch here at https://mdanalysisuserguide.readthedocs.io/en/rtd/

Also unsure how it works with directories just yet, but you could point to a subdomain and have userguide.mdanalysis.org/en/latest and userguide.mdanalysis.org/en/develop

richardjgowers:
So looks like using RTD allows us to build multiple branches at the "cost" of having a small advert in the sidebar? Seems like a good idea

richardjgowers:
Or wait, we can have the dev docs built on RTD with the current setup (gh-pages) for stable docs?

orbeckst:
> On Jun 7, 2020, at 8:15 AM, Richard Gowers <notifications@github.com> wrote:
> 
> 
> So looks like using RTD allows us to build multiple branches at the "cost" of having a small advert in the sidebar? Seems like a good idea
> 

[...]

PR #79, opened 2020-06-26
Doc versioning for the user guide

lilyminium:
Much as the documentation for the main repo, this PR implements manual doc versioning via Travis.

Demo: https://mdauserguide.minium.com.au/

- A PR merged into "master" or "develop" will generate docs (as a step towards #67)
- The script moves the HTML files into version subdirectories
- Sitemap index and versions.json file generated / updated
- **new** added redirect stubs to the root directory to point to the latest release, i.e.  https://mdauserguide.minium.com.au/trajectories/trajectories.html redirects to https://mdauserguide.minium.com.au/1.0.0/trajectories/trajectories.html -- it is much harder to implement a pagerule for this repo as the files aren't created in an easily matched `documentation_pages` directory
- most importantly doesn't wipe the CNAME every time a PR is merged


maintainer/deploy_docs_via_travis.sh
```diff
@@ -0,0 +1,84 @@
...
+ls *
+
+echo $(ls *.html | wc -l)
+
+if [ "$(ls *.html | wc -l)" -ge "1" ]; then
+    cd $VERSION && mdfiles=$(ls -d *) && cd -
+    for item in $mdfiles ; do
+        if [[ -f $item ]] || [[ -d $item ]]; then echo $item && git add $item; fi
+    done
+fi
```
orbeckst, line 77: This looks a bit fragile and I am not quite sure why it's needed after `git add -A  $VERSION`.

lilyminium: This adds the redirect files in the top level directory (`$mdfiles` is a list of all the files and directories inside `$VERSION`, and then we `cd -` to add the redirects). If we ``git add .`` it removes other `$VERSION` directories from previous commits, as it registers their absence. We could ignore removals but that doesn't help when we actually want to remove files. We could pull from the remote but I wasn't sure if this would result in merge conflicts. So this clumsy workaround.

I had to check if each `$item` exists because some directories don't contain `*.html` files, and apparently trying to `git add` a nonexistent item causes a fatal error and makes script exit.

orbeckst:
Thanks for the explanation. Maybe a short comment in the file for posterity would be helpfule (future selves/future maintainers)?

PR #81, opened 2020-06-28
Revert "Doc versioning for the user guide (#79)"

lilyminium:
This reverts commit 1b3063128553e2e9a161bc8350bad8cd9b59a684.


lilyminium:
Ok, this fails from segmentation faulting too, despite the Travis build for that commit passing earlier. Closing due to being unnecessary.

lilyminium:
Revert for now to see if we can fix the build with something we know worked before

Issue #82, opened 2020-06-29
Segmentation fault when building the user guide


lilyminium:
**Describe the bug**

Both RTD and Travis segmentation fault when building the user guide here:

```
reading sources... [ 98%] trajectories/transformations
Segmentation fault (core dumped)
```

Locally it works.

**To Reproduce**
1. Open a PR

**Symptoms**

- Pinning all the package versions did not fix it
- Just after reading all the sources, there are many warnings. Could these be related? 
```
UserGuide/doc/source/datasets.rst:6: WARNING: Duplicate explicit target name: "mdanalysisdata".
source/examples/analysis/alignment_and_rms/README.rst:18: WARNING: toctree contains reference to nonexisting document 'examples/analysis/alignment_and_rms/aligning_trajectory_to_first_frame'
source/examples/analysis/alignment_and_rms/README.rst:7: WARNING: duplicate label alignment-and-rms, other instance in UserGuide/doc/source/examples/analysis/alignment_and_rms/README.rst
source/examples/analysis/trajectory_similarity/README.rst:6: WARNING: duplicate label trajectory-similarity, other instance in UserGuide/doc/source/examples/analysis/trajectory_similarity/README.rst
UserGuide/doc/source/examples/analysis/alignment_and_rms/README.rst:18: WARNING: toctree contains reference to nonexisting document 'examples/analysis/alignment_and_rms/aligning_trajectory_to_first_frame'
UserGuide/doc/source/examples/analysis/alignment_and_rms/README.rst:7: WARNING: duplicate label alignment-and-rms, other instance in UserGuide/doc/source/examples/analysis/README.rst
UserGuide/doc/source/examples/analysis/trajectory_similarity/README.rst:6: WARNING: duplicate label trajectory-similarity, other instance in UserGuide/doc/source/examples/analysis/README.rst
UserGuide/doc/source/formats/coordinates.rst:4: WARNING: duplicate label coordinate-readers, other instance in UserGuide/doc/source/formats/index.rst
source/formats/coordinates.rst:4: WARNING: duplicate label coordinate-readers, other instance in UserGuide/doc/source/formats/coordinates.rst
UserGuide/doc/source/formats/reference/trj.rst:35: WARNING: Bullet list ends without a blank line; unexpected unindent.
looking for now-outdated files... none found
pickling environment... done
[...]

PR #91, opened 2020-07-08
Try conda environment

lilyminium:
Try different things relating to #82.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,10 @@ env:
   - SPHINX_DIR=doc/
   - HTML_DIR=doc/build/html
   - PYTHON_VERSION=3.6
-  - CONDA_DEPENDENCIES="jupyter_contrib_nbextensions nglview sphinx ipywidgets MDAnalysis MDAnalysisTests plotly"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -18,11 +18,12 @@ env:
   - SPHINX_DIR=doc/
   - HTML_DIR=doc/build/html
   - PYTHON_VERSION=3.6
```
kain88-de, line 20: try python version 3.8

lilyminium:
Thanks @kain88-de, I see you’ve done that :-) the previous builds were actually failing due to astropy’s ci-helpers hanging after building from the environment.yml, wasn’t sure if that was a bug on their, ours, or the Travis end. Any ideas?

kain88-de:
Conda tripping over itself. Hard to say why.

kain88-de:
I have no idea why there is a segfault. It is also really hard to see where the code has problems because sphinx is not very verbose here. No idea what is happening. But I do not like that our conda packages segfault.

lilyminium:
Closing due to now having to work out how to do this via GH-actions.

PR #83, opened 2020-06-29
Re-version docs

lilyminium:
- Updated version of #79, please see for comments
- in #81 the commit was reverted to focus on fixing #82: segmentation fault when building the user guide
- Supercedes #80 (which was merged into this)

PR #75, opened 2020-06-14
Update rms notebooks

lilyminium:
RMS notebook updated for 1.0:

* added `weights_groupselections`
* updated charge-weighting example (a bit meaningless as CAs all had same charge) to residue-number weighting, with emphasis that it's an **example only**

Built: https://mdauserguide.readthedocs.io/en/rms-1.0/

Notebooks:

* https://mdauserguide.readthedocs.io/en/rms-1.0/examples/analysis/alignment_and_rms/rmsd.html

* https://mdauserguide.readthedocs.io/en/rms-1.0/examples/analysis/alignment_and_rms/pairwise_rmsd.html

PR #95, opened 2020-07-18
pin numpy to 1.18.5

IAlibay:
So I was testing what could be causing this, and given the timelines and everything it seems like it _could_ be down to numpy 1.19.

At least in my own test case (I created a clean repo with a call to aligntraj with a u.copy()), it seems that this is the case, let's see how it fares here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
 
 install:
   - conda env create python=3.6 -f environment.yml
-  - conda activate mda-user-guide
+  - source activate mda-user-guide
   - jupyter-nbextension enable nglview --py --sys-prefix
   - conda list
   
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
   - SPHINX_DIR=doc/
   - HTML_DIR=doc/build/html
   - PYTHON_VERSION=3.6
-  - CONDA_DEPENDENCIES="numpy==1.18.5 jupyter_contrib_nbextensions nglview sphinx ipywidgets MDAnalysis MDAnalysisTests plotly"
+  - CONDA_DEPENDENCIES="jupyter_contrib_nbextensions nglview sphinx ipywidgets MDAnalysis MDAnalysisTests plotly"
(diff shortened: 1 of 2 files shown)


IAlibay:
Summary:
1.18.5 = a5928a6, CI passes
numpy version not defined = 955f6f4, CI fails
1.19 = 87d8b85, CI fails

The failure seen here isn't the same as #75, but it does seem like it might partly be the cause.

lilyminium:
Thank you for finding this @IAlibay! Given that [this commit](https://github.com/MDAnalysis/UserGuide/pull/83/commits/f722bcf78e3bbd657fcd1895cf754263789623a1) failed without astropy ci-helpers, could you please see if it works here as well?

lilyminium:
> $ conda install --yes numpy<1.19.0

> /home/travis/.travis/functions: line 109: 1.19.0: No such file or directory

> The command "conda install --yes numpy<1.19.0" failed and exited with 1 during .

You probably need to put that in quotes, `conda install -y 'numpy<1.19.0'`.

Looking at the passing commit, it seems that `numpy` and `numpy-base` are *not* from `conda-forge` as we thought but instead, "blank" (I'm guessing defaults) -- what if you try putting defaults::numpy-base into the environment.yml? I'm not sure how that then works with scipy then, which I thought needed conda-forge packages...

orbeckst:
Random comment from @richardjgowers https://github.com/MDAnalysis/pytng/pull/29#issuecomment-661733926 (my highlighting)
> A lot of methods you find won't work because of the memory being allocated by malloc/realloc/free, whereas memory in numpy is (by default) handled using PyMem_malloc which is a different system and **mixing the two causes segfaults**. I.e. you can't malloc then PyMem_Free. This is all to avoid doing copies of data, but I'd really like to keep the bindings tight else we're letting everything downstream down.

Is it possible that our Cython-based XDRReader does something somewhere where we are not careful with how we pass arrays around (and maybe something changed in that respect in numpy recently?)

Sorry, grasping at straws here...

lilyminium:
Closing due to now having to work out how to do this via GH-actions.

PR #101, opened 2020-08-16
Add examples of selecting groups from other groups

mieczyslaw:
Although, both syntactically correct, only one solution to every use case returns expected results. This should help people to avoid using AtomGroups incorrectly, especially when seeing all the possibilities via a interactive debugging session on objects.

I have rebuilt the documentation and confirmed that the webpage after updates is as expected.

### DIFF
--- a/doc/source/groups_of_atoms.rst
+++ b/doc/source/groups_of_atoms.rst
@@ -66,7 +66,8 @@ For information on adding custom Residues or Segments, have a look at :ref:`addi
 
 Access to other classes via the :class:`~MDAnalysis.core.groups.AtomGroup` object can be pretty powerful, but also
 needs to be used with caution to avoid accessing data outside the intended selection. Therefore, we present two use
-cases showing commonly used applications on a simple extract from the PDB file:
+cases showing commonly used applications, for which we define :class:`~MDAnalysis.core.universe.Universe`
+on a simple extract from the PDB file:
 
 .. ipython:: python
 
@@ -102,6 +103,7 @@ cases showing commonly used applications on a simple extract from the PDB file:
     ATOM   1045  HA2 GLY B 123     -26.430  23.831 -11.600  1.00  0.00           H
     TER
     """)
+    u = mda.Universe(pdb, format="PDB")
 
 Use case: Sequence of residues by segment
 -----------------------------------------
@@ -109,26 +111,27 @@ In order to select only ATOM record types and get a list of residues by segment,
 
 .. ipython:: python
 
-    u = mda.Universe(pdb, format="PDB")
-    residues = list()
+    residues_by_seg = list()
     for seg in u.segments:
         p_seg = seg.atoms.select_atoms("record_type ATOM")
-        residues.append(p_seg.residues.resnames)
-    residues
+        residues_by_seg.append(p_seg.residues)
 
-As expected, HETATM record type lines are not considered.
+Residue names can be extracted using Python's list comprehensions. As required, HETATM record type lines are not
+considered:
 
-Although syntactically correct, accessing residues via segments (to avoid looping through segments),
-would lead to **unexpected** output
+.. ipython:: python
+
+    [rg.resnames for rg in residues_by_seg]
+
+Note that accessing residues by first selecting the segments of an :class:`~MDAnalysis.core.groups.AtomGroup` returns
+all the residues in that segment for both the ATOM and HETATM record types (no memory of the original selection).
+The meaning of this is: "give me all residue names from segments in which there is at least one of the selected atoms".
 
 .. ipython:: python
 
     selected_atoms = u.select_atoms("record_type ATOM")
-    all_residues = selected_atoms.segments.resnames
-    all_residues
-
-containing residues both for ATOM and HETATM record types; this construct means
-"give me all residue names from segments in which there is at least one of the selected atoms".
+    all_residues = selected_atoms.segments.residues
+    all_residues.resnames
 
 Use case: Atoms list grouped by residues
 ----------------------------------------
@@ -136,31 +139,38 @@ In order to list all the heavy protein backbone and sidechain atoms in every res
 
 .. ipython:: python
 
-    u = mda.Universe(pdb, format="PDB")
     atoms_in_residues = list()
     for seg in u.segments:
         p_seg = seg.atoms.select_atoms("record_type ATOM and not name H*")
         for p_res in p_seg.residues:
-            atoms_in_residues.append(p_seg.select_atoms(f"resid {p_res.resid} and resname {p_res.resname}").names)
-    atoms_in_residues
+            atoms_in_residues.append(p_seg.select_atoms(f"resid {p_res.resid} and resname {p_res.resname}"))
 
-As expected, HETATM record type lines and hydrogen atoms are not considered.
+Atom names can be extracted using Python's list comprehensions. As required, HETATM record type lines and hydrogen
+atoms are not considered:
 
-Although syntactically correct, accessing atoms via residues (to avoid looping through residues),
-would lead to **unexpected** output
+.. ipython:: python
+
+    [ag.names for ag in atoms_in_residues]
+
+The Python syntax can be further simplified by using ``split()`` function:
+
+.. ipython:: python
+
+    rds = u.select_atoms("record_type ATOM and not name H*").split("residue")
+    [ag.names for ag in rds]
+
+Note that accessing atoms by first selecting the residues of an :class:`~MDAnalysis.core.groups.AtomGroup` also
+returns hydrogen atoms (no memory of the original selection). The meaning of this is "give me all atom names from
+residues in which there is at least one of the selected atoms". However, it doesn't contain a nitrogen atom from XYZ
+residue as no atoms from this residue were in the :class:`~MDAnalysis.core.groups.AtomGroup`.
 
 .. ipython:: python
 
     all_atoms_in_residues = list()
     for seg in u.segments:
         p_seg = seg.atoms.select_atoms("record_type ATOM and not name H*")
-        all_atoms_in_residues.append(p_seg.residues.names)
-    all_atoms_in_residues
-
-containing hydrogen atoms; this construct means "give me all atom names from residues
-in which there is at least one of the selected atoms". Please note that it doesn't contain a
-nitrogen atom from XYZ residue as no atoms from this residue were in the
-:class:`~MDAnalysis.core.groups.AtomGroup`.
+        all_atoms_in_residues.append(p_seg.residues)
+    [atom.names for atom in all_atoms_in_residues]
 
 ---------------------------
 Fragments
--- a/doc/source/groups_of_atoms.rst
+++ b/doc/source/groups_of_atoms.rst
@@ -64,65 +64,103 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the 
 
 For information on adding custom Residues or Segments, have a look at :ref:`adding-residue`.
 
-Access to various data views via :class:`~MDAnalysis.core.groups.AtomGroup` object can be pretty powerful, but also
-needs to be used with caution to avoid accessing data outside the intended selection, as mentioned above. Therefore,
-we present two use cases showing commonly used applications.
+Access to other classes via the :class:`~MDAnalysis.core.groups.AtomGroup` object can be pretty powerful, but also
+needs to be used with caution to avoid accessing data outside the intended selection. Therefore, we present two use
+cases showing commonly used applications on a simple extract from the PDB file:
 
----------------------------------------
-Use case: Sequence of residues by chain
----------------------------------------
-Let's assume that we have a PDB file, which contains both ATOM and HETATM record types, for protein and small molecule,
-respectively. In order to select only ATOM record types and get a list of residues by chain, one needs to call:
+.. ipython:: python
 
-.. code-block:: python
+    import MDAnalysis as mda
+    import io
+    pdb = io.StringIO("""
+    ATOM    414  N   GLY A 402     -51.919   9.578 -14.287  1.00 68.46           N
+    ATOM    415  CA  GLY A 402     -52.405  10.954 -14.168  1.00 68.41           C
+    ATOM    416  C   GLY A 402     -51.821  11.946 -15.164  1.00 69.71           C
+    ATOM    417  O   GLY A 402     -51.958  13.159 -14.968  1.00 69.61           O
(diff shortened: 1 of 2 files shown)


doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,66 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+
+.. code-block:: python
+
+    u = mda.Universe(pdb, format="PDB")
+    for seg in u.segments:
+        p_seg = seg.atoms.select_atoms("record_type ATOM")
+        r_list = p_seg.residues.resnames
+        print(r_list)
+
+This can be used later on to check whether the residue names belong, for example, to the list of standard residues.
```
lilyminium, line 85: This doesn't seem like a very targeted use case for splitting the residue names by segment. I would either come up with a more specific scenario or delete this line.

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,66 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+However, it would be **incorrect** to use:
+
+.. code-block:: python
+
+    selected_atoms = u.select_atoms("record_type ATOM")  # CORRECT SYNTAX, BUT INCORRECT RETURN VALUE!
```
lilyminium, line 90: Can you please explain why this is an incorrect return value? Phrasing it this way makes it sound like a bug, but it's actually a reasonable step towards `u.select_atoms("record_type ATOM").split("segment")`, which is another way to do this.

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,66 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+    residues_wrong = selected_atoms.segments.resnames  # CORRECT SYNTAX, BUT INCORRECT RETURN VALUE!
+    print(residues_wrong)  # CORRECT SYNTAX, BUT INCORRECT RETURN VALUE!
```
lilyminium, line 92: You probably don't need to comment "# CORRECT SYNTAX, BUT INCORRECT RETURN VALUE!" on every line; it would be enough to simply explain what is going on and why you might prefer something else, in text that comes in between each line. If you show the output, that is also immediately clearer.

(Also, I find the all-caps error message with exclamation mark quite alarming 😅)

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,66 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+
+because, although it nicely returns list of residues in every chain (list of lists, so no need to additionally
+loop through segments), it would return all residues, both for ATOM and HETATM record types, because this construct
+means "give me all residue names from segments in which there is at least one of the selected atoms".
```
lilyminium, line 96: Perhaps break this up into smaller sentences and intersperse the code with it, explaining what each line does.

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,66 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+However, it would be **incorrect** to use in the body of the first loop:
+
+.. code-block:: python
+
+    atoms_in_residue_wrong = p_seg.residues.names  # CORRECT SYNTAX, BUT INCORRECT RETURN VALUE!
+    print(atoms_in_residue_wrong)  # CORRECT SYNTAX, BUT INCORRECT RETURN VALUE!
+
+because, although it nicely returns list of atoms in every residue, it would return all atoms (including hydrogen ones),
+because this construct means "give me all atom names from residues in which there is at least one of the selected
+atoms".
```
lilyminium, line 125: This scenario is a bit more involved and I think would be better as a separate, small tutorial, for people who do want to do this task. You could explain why you do not want the "H*" and "OXT" atoms, show where you can get a list of atoms in standard residues, and show the comparison via symmetric difference. You could then link to that tutorial here.

Otherwise this use case assumes a fair bit biological/MD background knowledge (e.g. why avoid HETATM? what is a HETATM vs ATOM?)

Side note, I think you could also do (untested):

```python
segs = u.select_atoms("record_type ATOM and not name H* OXT").split("residue")
print([ag.names for ag in segs])
```

If I recall correctly, residues are always in one segment or another, they cannot be split over segments.

But it is worthwhile going over loops etc. for new users.

mieczyslaw:
Thanks @lilyminium for reviewing this work so quickly! I hope to address all suggestions this week.

doc/source/groups_of_atoms.rst
orbeckst: If it's a self-contained example then it could become a notebook in the [Examples](https://userguide.mdanalysis.org/1.0.0/examples/README.html) section, couldn't it?

mieczyslaw: Excellent idea @orbeckst ! Where is this Examples page you think about? Do you mean adding it here: https://www.mdanalysis.org/MDAnalysisTutorial/atomgroups.html ?

I can add those to Examples and reference from AtomGroup page. Is that OK to add a few lines of PDB file to the example rather than adding it to `MDAnalysis.tests.datafiles` ?

orbeckst: No, the MDAnalysisTutorial is outdated and really deprecated.

I mean the Examples section in the User Guide https://userguide.mdanalysis.org/examples/README.html – add a new page there.

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,104 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+"give me all residue names from segments in which there is at least one of the selected atoms".
+
+Use case: Atoms list grouped by residues
+----------------------------------------
+In order to list all the heavy protein backbone and sidechain atoms in every residue, one needs to call:
+
+.. ipython:: python
+
+    u = mda.Universe(pdb, format="PDB")
+    atoms_in_residues = list()
```
orbeckst, line 140: Collect Atoms (ie AtomGroups) not names.

Just collecting strings does not really show how to work with groups. Just extract the information from the group at the end. 

I have seen users parse `__repr__` with regex to get the information that's in `.resnames`, `.names` etc... so the more we can show how to get this information in a sensible manner the better.

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,104 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+    for seg in u.segments:
```
orbeckst, line 141: Version with `split()`?

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,104 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+        p_seg = seg.atoms.select_atoms("record_type ATOM and not name H*")
+        for p_res in p_seg.residues:
+            atoms_in_residues.append(p_seg.select_atoms(f"resid {p_res.resid} and resname {p_res.resname}").names)
+    atoms_in_residues
+
+As expected, HETATM record type lines and hydrogen atoms are not considered.
+
+Although syntactically correct, accessing atoms via residues (to avoid looping through residues),
```
orbeckst, line 149: See comment above on tone.

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,104 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+
+.. ipython:: python
+
+    all_atoms_in_residues = list()
+    for seg in u.segments:
+        p_seg = seg.atoms.select_atoms("record_type ATOM and not name H*")
+        all_atoms_in_residues.append(p_seg.residues.names)
+    all_atoms_in_residues
+
+containing hydrogen atoms; this construct means "give me all atom names from residues
```
orbeckst, line 160: Explanation is good.

"please" is not needed here, IMO — it doesn't fit the overall writing style of the UG

orbeckst (changes requested):
Thank you for clarifying use of groups. My three major requests are
- return groups and not strings; this is a section about groups; just show at the end how to extract the strings (using list comprehensions is fine)
- add examples with `split()` which is really our recommended way to deal with breaking up groups; the explicit examples can definitely be shown but then demonstrate how this can be written more concisely (as @lilyminium suggested)
- change the tone of the sections that were written from the initial user perspective of "unexpected" behavior: you are now writing from the perspective of the API and User Guide, which does not see anything unexpected there. If you think that some users (not all) might find it unexpected then you need to address it in a more positive manner (see inline comments). Ultimately @lilyminium has the last word on the tone as she has been shaping the User Guide from the beginning.

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,104 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+    u = mda.Universe(pdb, format="PDB")
+    residues = list()
+    for seg in u.segments:
+        p_seg = seg.atoms.select_atoms("record_type ATOM")
+        residues.append(p_seg.residues.resnames)
+    residues
+
+As expected, HETATM record type lines are not considered.
+
+Although syntactically correct, accessing residues via segments (to avoid looping through segments),
```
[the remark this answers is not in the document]
lilyminium: I agree with @orbeckst, and personally do not like to make assumptions about what users might expect. You could instead say something like "Accessing residues by first selecting the segments of an `AtomGroup` returns all the residues in that segment for both the ATOM and HETATM record types." and explain why (`selected_atoms.segments` gives all the segments in `selected_atoms`; `selected_atoms.segments.residues` gives all the residues in those segments, with no memory of the original selection). That would clarify some of the basics laid down higher on the page about the hierarchy of the atom groupings in a really nice way!

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,104 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+would lead to **unexpected** output
+
+.. ipython:: python
+
+    selected_atoms = u.select_atoms("record_type ATOM")
+    all_residues = selected_atoms.segments.resnames
+    all_residues
```
lilyminium, line 128:
```suggestion
    all_residues = selected_atoms.segments.residues
    all_residues.resnames
```

I think this distinction would make the organisation of the groups clearer.

lilyminium (changes requested):

doc/source/groups_of_atoms.rst
lilyminium: > I have seen users parse __repr__ with regex to get the information that's in .resnames, .names etc... so the more we can show how to get this information in a sensible manner the better.

That's impressive... do topology attributes show up in `dir()`?

lilyminium:
Sorry for falling out of touch with this PR, @mieczyslaw and @orbeckst -- work has been a bit busy lately.

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,104 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+    TER
+    """)
+
+Use case: Sequence of residues by segment
+-----------------------------------------
+In order to select only ATOM record types and get a list of residues by segment, one needs to call:
+
+.. ipython:: python
+
+    u = mda.Universe(pdb, format="PDB")
```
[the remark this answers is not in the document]
mieczyslaw: Done

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,104 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+    residues = list()
```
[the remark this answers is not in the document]
mieczyslaw: Done

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,104 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+    for seg in u.segments:
+        p_seg = seg.atoms.select_atoms("record_type ATOM")
+        residues.append(p_seg.residues.resnames)
```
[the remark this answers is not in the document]
mieczyslaw: Done, without print() as iPython outputs the results.

doc/source/groups_of_atoms.rst
mieczyslaw: Done

doc/source/groups_of_atoms.rst
mieczyslaw: Done

doc/source/groups_of_atoms.rst
mieczyslaw: Done

doc/source/groups_of_atoms.rst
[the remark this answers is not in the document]
mieczyslaw: Done. Hope it is now improved to the required standard.

doc/source/groups_of_atoms.rst
mieczyslaw: Done. Hope it is now improved to the required standard.

doc/source/groups_of_atoms.rst
```diff
@@ -64,6 +64,104 @@ Similarly, an :class:`~MDAnalysis.core.groups.Atom` has direct knowledge of the
...
+
+Use case: Sequence of residues by segment
+-----------------------------------------
+In order to select only ATOM record types and get a list of residues by segment, one needs to call:
+
+.. ipython:: python
+
+    u = mda.Universe(pdb, format="PDB")
+    residues = list()
+    for seg in u.segments:
```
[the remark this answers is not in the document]
mieczyslaw: Done

doc/source/groups_of_atoms.rst
mieczyslaw: yes, this works with split()

mieczyslaw:
I did my best to address all of your comments, @lilyminium and @orbeckst . Please kindly review and let me know if there are anyc changes you'd like me to make. Thanks.

orbeckst (approved):

lilyminium (approved):

lilyminium:
This looks great, thank you for adding these clarifying examples @mieczyslaw!

PR #104, opened 2020-09-16
Turn on scripts

lilyminium:
Run the scripts that generate format pages / topology attribute tables on building docs. This was previously turned off for 0.20.1 because the package itself was missing the infrastructure needed.


doc/source/conf.py
```diff
@@ -64,7 +64,7 @@ def sort_authors(filename):
 # TURNED OFF FOR NOW until we sort out how versioned docs
 # work and how we will install MDAnalysis + dependencies
 #
-# subprocess.call('./scripts/generate_all.sh')
```
orbeckst, line 67: presumably the comments can go when it is enabled?

.gitignore
```diff
@@ -22,8 +22,8 @@ doc/build
 *.so
 
 # Ignore generated docs
-# */**/*.txt
-# !requirements.txt
+*/**/*.txt
```
orbeckst, line 25: Not sure if this is working because there are some *.txt files in this PR... or maybe they are in a different place?

doc/source/formats/topology_parsers.txt
```diff
@@ -3,30 +3,30 @@
 
 .. table:: Table of supported topology parsers and the attributes read
 
-    ======================================  ====================================================================  ===================================================================================================================================================================================================================  ====================
```
orbeckst, line 6: Should this *.txt file have been ignored?

doc/source/scripts/gen_topologyparser_attrs.py
```diff
@@ -8,54 +8,53 @@
...
 
-parser_attrs = {}
 
-for p in PARSER_TESTS:
-    e, g = set(p.expected_attrs)-MANDATORY_ATTRS, set(p.guessed_attrs)
-    # clunky hack for PDB
-    if p is TestPDBParser:
-        e.add('elements')
-    parser_attrs[p.parser] = (e, g)
+def collect_classes():
```
orbeckst, line 26: yes, but still better than hardcoding  and missing something

orbeckst (commented):
I am not really qualified to review this but because I looked out of curiosity I left comments.

doc/source/formats/topology_parsers.txt
lilyminium: I think I may have pushed changes before adding it to `.gitignore`.

PR #110, opened 2020-10-21
Add stable/ and dev/ versions of UserGuide

lilyminium:
Parallel to #2968 and #2970, this creates copies of 1.0.0 to stable/ and 2.0.0-dev0 dev/ versions of the User Guide.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,6 @@ after_success:
   - |
     if [[ ${TRAVIS_PULL_REQUEST} == "false" ]]; then
       python -c 'import os,sys,fcntl; flags = fcntl.fcntl(sys.stdout, fcntl.F_GETFL); fcntl.fcntl(sys.stdout, fcntl.F_SETFL, flags&~os.O_NONBLOCK);'
-      export URL="https://userguide.mdanalysis.org/"
       export VERSION=$(python -c "import MDAnalysis; print(MDAnalysis.__version__)")
       export MAINTAIN_DIR=${TRAVIS_BUILD_DIR}/maintainer
       export MDA_DOCDIR=$HTML_DIR
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ os:
 
 env:
   global:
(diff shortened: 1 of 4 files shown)


maintainer/update_json_stubs_sitemap.py
```diff
@@ -107,10 +108,66 @@ def write_redirect(file, version='', outfile=None):
...
+    print(f"Redirected URLs in {file} from {old} to {new}")
+
+
+def add_or_update_version(version):
+    """Add or update the version path to versions.json"""
+    for ver in versions:
+        if ver["version"] == version:
+            ver["url"] = os.path.join(URL, version)
+            break
+        else:
```
orbeckst, line 146: Should this `else` correspond to `for` to avoid the "multiple dev" problem?

.travis.yml
```diff
@@ -43,7 +43,6 @@ after_success:
   - |
     if [[ ${TRAVIS_PULL_REQUEST} == "false" ]]; then
       python -c 'import os,sys,fcntl; flags = fcntl.fcntl(sys.stdout, fcntl.F_GETFL); fcntl.fcntl(sys.stdout, fcntl.F_SETFL, flags&~os.O_NONBLOCK);'
-      export URL="https://userguide.mdanalysis.org/"
```
orbeckst, line 46: Do you prefer to set the URL in the script instead of in the travis settings? You could add it to the global variables at the top of travis.yml as one of the major configuration settings. 

There are also still all the other variables that remain in the script. Why is URL moved?

orbeckst (changes requested):
One potential fix (`for ... else`?) and question regarding `URL`.

.travis.yml
lilyminium: Probably just absentmindedness. This now needs to be replaced with something similar to the gh-actions pipeline in mdanalysis proper (https://github.com/MDAnalysis/mdanalysis/pull/3051)

lilyminium:
We should replace this with GH-actions.

PR #115, opened 2020-12-06
Install develop version of MDAnalysis in develop version of user guide

lilyminium:
We need to install the development version of MDAnalysis in order to run the ``ipython`` code that relies on the new changes.


lilyminium:
Closed as now we use different versions of `environment.yml` that will handle the MDA versioning for us.

PR #135, opened 2021-02-12
Use RTD to make PDF of the docs

lilyminium:
As requested on Discord.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,6 +5,11 @@
 # Required
 version: 2
 
+formats:
+  - epub
+  - pdf
+  - htmlzip
+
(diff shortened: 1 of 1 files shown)


jbarnoud:
I looked at the RTD build and could not find the epub or PDF versions. Where are they supposed to appear?

lilyminium:
I think it doesn't build PDFs for pull requests, so I might merge this to see if it does for master -- it should be in the [Downloads](https://readthedocs.org/projects/mdanalysisuserguide/downloads/) area.

lilyminium:
@jbarnoud it now appears in the [Downloads](https://readthedocs.org/projects/mdanalysisuserguide/downloads/) 🍡

PR #203, opened 2022-04-18
Improve development environment installation instructions

RMeli:
This PR is an attempt to improve the installation instructions of an MDA developement environment based on questions and errors seen on the GSoC and Outreachy Discord channels.

The main changes are the following:
* Added a note to use `conda` for first time contributors (this should simplify things on Windows, from what I gathered)
* Add warning to bring attention to the fact that an environment should be active when developing MDA
* Move deactivation instructions into a note block (this avoids unintentional deactivation, if instructions are followed by copy-pasting the code blocks)
* Split conda and pip/venv instructions in "Build MDA" section to avoid confusion
  * Added note to make sure the correct environment is activated
* Remove installations with Linux-specific package managers


doc/source/contributing_code.rst
```diff
@@ -213,18 +216,48 @@ and activated your virtual environment. First we need to install dependencies. I
             matplotlib mmtf-python mock netcdf4 networkx \
             "numpy>=1.17.3" psutil pytest scikit-learn scipy \
```
IAlibay, line 217:
```suggestion
            "numpy>=1.18.0" psutil pytest scikit-learn scipy \
```

doc/source/contributing_code.rst
```diff
@@ -213,18 +216,48 @@ and activated your virtual environment. First we need to install dependencies. I
...
             "seaborn>=0.7.0,<0.9" sphinx==1.8.5 "tidynamics>=1.0.0" \
```
IAlibay, line 218: Not sure where the <0.9 for seaborn came from, maybe py2.7? Anyways, CI doesn't have a max limit.
We got rid of the sphinx limit, so we don't have to limit it to 1.8.5 anymore.

```suggestion
            "seaborn>=0.7.0" sphinx "tidynamics>=1.0.0" \
```

doc/source/contributing_code.rst
```diff
@@ -202,8 +199,14 @@ This sets the number of files to 4096. However, this command only applies to you
...
+With conda
+----------
+
+.. note::
+    Make sure that you have :ref:`cloned the repository <forking-code-repo>`  
+    and activated your virtual environment with :code:`conda activate mdanalysis-dev`.
+    
+First we need to install dependencies. You'll need a mix of conda and pip installations:
 
     .. code-block:: bash
```
IAlibay, line 212: I think we can make the switch from biobuilds to bioconda as long as we're happy that the latter is suitably secure/well-used? (which I think it is)

doc/source/contributing_code.rst
```diff
@@ -235,6 +268,8 @@ If you're using pip, it is a little simpler. However, some packages such as ``cl
           sphinx==1.8.5 sphinx_rtd_theme "tidynamics>=1.0.0" \
```
IAlibay, line 268:
```suggestion
          sphinx sphinx_rtd_theme "tidynamics>=1.0.0" \
```

doc/source/contributing_code.rst
```diff
@@ -213,18 +216,48 @@ and activated your virtual environment. First we need to install dependencies. I
...
+-----------------------
+
+.. note::
+    Make sure that you have :ref:`cloned the repository <forking-code-repo>`  
+    and activated your virtual environment with :code:`source myproject-env/bin/activate`
+    (or :code:`workon my-project` if you used the `virtualenvwrapper package <https://virtualenvwrapper.readthedocs.io/en/latest/>`_)
+
+Install the dependencies:
 
     .. code-block:: bash
```
IAlibay, line 263: As above for seaborn and numpy min/max versions.

doc/source/generated/selections/base.txt
```diff
@@ -1,7 +1,7 @@
 ..
     Generated by gen_standard_selections.py
 
-==  ==  ==  ==  ==  ==  ==  ===
```
IAlibay, line 4: It's not clear from your PR message as to why these are all changing?

IAlibay (changes requested):
Couple of quick comments.

doc/source/generated/selections/base.txt
RMeli: Not sure TBH. They change when I build the documentation, so I committed them since they are not ignored. They will likely be overwritten when the docs are build in CI?

IAlibay: 🤷🏽 maybe @lilyminium knows

lilyminium: Ah yeah, sorry, I think I can't ignore them because the docs are built in CI? Is there a way to ignore files only from a user commit and not from a GH-Actions commit?

But yeah they get rebuilt.

doc/source/contributing_code.rst
```diff
@@ -60,6 +60,10 @@ both modules in development mode.
 
 You can do this either with :ref:`conda <dev-with-conda>` or :ref:`pip <dev-with-pip>`.
 
+.. note::
+    If you are a first time contributor and/or don't have a lot of experience managing
+    your own Python virtual environments, we suggest to use :code:`conda`.
```
lilyminium, line 65:
```suggestion
    your own Python virtual environments, we **strongly** suggest using :code:`conda`.
```

*Really really strongly* :p

doc/source/contributing_code.rst
```diff
@@ -202,39 +199,77 @@ This sets the number of files to 4096. However, this command only applies to you
...
+        pip install -e .
+
+At this point you should be able to import MDAnalysis from your locally built version. If you are running the development version, this is visible from the version number ending in :code:`-dev0`. For example:
+
+    .. code-block:: bash
+
+        $ python  # start an interpreter
+        >>> import MDAnalysis as mda
+        >>> mda.__version__
+        '2.1.0-dev0'
```
lilyminium, line 246: Would `python -c "import MDAnalysis as mda; print(mda.__version__)"` be more foolproof?

doc/source/generated/selections/base.txt
lilyminium: The changes are odd though. I think this means there's already txt files in the `develop` branch that we should probably remove. The ones in gh-pages (at least rendered at the https://userguide.mdanalysis.org/stable/units.html) look up to date with Rocco's, so that's not too bad.

lilyminium (commented):
Thanks, this looks like a solid improvement. I'd put more in about using Windows specifically, but I don't know how to use Windows ☹️

lilyminium:
Are you already in AUTHORS?

doc/source/contributing_code.rst
```diff
@@ -60,6 +60,10 @@ both modules in development mode.
 
 You can do this either with :ref:`conda <dev-with-conda>` or :ref:`pip <dev-with-pip>`.
```
lilyminium, line 61: Actually if you could add a note about only using **one** of them, I think a few people tried doing both?

doc/source/generated/selections/base.txt
RMeli: Ok, so it is safe to leave them?

doc/source/contributing_code.rst
RMeli: Yep, I like it!

doc/source/contributing_code.rst
RMeli: Yeah, that's why I split the two instructions when building MDA but making is extra clear is great.

doc/source/contributing_code.rst
```diff
@@ -114,15 +119,11 @@ With pip and virtualenv
 
 Like conda, virtual environments managed with `virtualenv <https://virtualenv.pypa.io/en/latest/>`_ allow you to use different versions of Python and Python packages for your different project. Unlike conda, virtualenv is not a general-purpose package manager. Instead, it leverages what is available on your system, and lets you install Python packages using pip.
 
-To use virtual environments you have to install the virtualenv package first. This can be done with either pip or the package manager of your system:
+To use virtual environments you have to install the virtualenv package first. This can be done with pip:
 
     .. code-block:: bash
 
         pip install virtualenv
```
IAlibay, line 126: Here and everywhere else, I remember someone telling me that using `python -m pip` was actually safer than relying on pip having been picked up properly by your environment. Thoughts on switching everything over to that?

doc/source/contributing_code.rst
```diff
@@ -202,39 +199,77 @@ This sets the number of files to 4096. However, this command only applies to you
...
+    (or :code:`workon my-project` if you used the `virtualenvwrapper package <https://virtualenvwrapper.readthedocs.io/en/latest/>`_)
+
+Install the dependencies:
 
     .. code-block:: bash
 
         pip install biopython chemfiles codecov cython \
           griddataformats gsd hypothesis "joblib>=0.12" matplotlib \
           msmb_theme==1.2.0 netcdf4 networkx "numpy>=1.17.3" \
           parmed psutil pytest scikit-learn scipy "seaborn>=0.7.0,<0.9" \
```
IAlibay, line 267:
```suggestion
          msmb_theme==1.2.0 netcdf4 networkx "numpy>=1.18.0" \
          parmed psutil pytest scikit-learn scipy "seaborn>=0.7.0" \
```

doc/source/contributing_code.rst
```diff
@@ -202,39 +199,77 @@ This sets the number of files to 4096. However, this command only applies to you
...
-            "tqdm>=4.43.0"
-
-        # if using conda with python 3.7 or 3.8, also run
-        conda install -c conda-forge parmed
-
-        # if using conda with other versions of python, also run
-        pip install parmed
+            "numpy>=1.18.0" psutil pytest scikit-learn scipy \
+            "seaborn>=0.7.0" sphinx "tidynamics>=1.0.0" \
+            "tqdm>=4.43.0" parmed
```
IAlibay, line 219: @lilyminium thoughts on also adding openmm and rdkit here?

I'm kinda reluctant to get folks to pull down openmm without the cuda shim though...

IAlibay (changes requested):
Couple of other small things.

doc/source/contributing_code.rst
RMeli: Good point. I think if you did not install it explicitly with `conda` you can pick up the system `pip` or the `base` environment one. I'd be in favour of switching (that's how I use it), unsless there are drawbacks I'm unaware of.

doc/source/contributing_code.rst
lilyminium: I have nothing against rdkit but would be mostly against installing openmm unless absolutely needed, which surely the user would install if they were using the converter function to have an OpenMM object to convert. Seems quite big

lilyminium: Honestly, even parmed is probably unnecessary here?

doc/source/generated/selections/base.txt
lilyminium: Yeah -- or you could remove them completely, doesn't make much sense to keep txt files in the repo. Not much difference either way. Maybe you don't want your path commented in the tables though?

doc/source/contributing_code.rst
IAlibay: We could drop all the converter dependencies and note it somewhere in text?

lilyminium: Sure, I don't think the text is even super essential though -- you need the library installed to convert to MDAnalysis, and the converters from MDAnalysis have their own error messages telling you to install them.

IAlibay: Sounds good - let's remove parmed. Folks usually won't play around much with converters unless they know what they are doing anyways.

doc/source/contributing_code.rst
IAlibay: Unless @lilyminium has any objections I say let's do it.

doc/source/contributing_code.rst
RMeli: Sounds good. I'll push the changes tonight.

doc/source/contributing_code.rst
RMeli: Given that we _strongly_ recommend `conda`, should we remove the information about `virtualenvwrapper`? If someone wants to go down the `virtualenv` route, probably they know what they are doing and know about it already?

doc/source/generated/selections/base.txt
RMeli: Good point. So I can remove all files in `generated`?

lilyminium: I think so yeah

RMeli: How can I re-generate these files? If I run `make html` after removing them I get a bunch of errors:
```
[...]
/Users/rmeli/Documents/git/software/UserGuide/doc/source/topology_system.rst:152: CRITICAL: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: 'source/generated/topology/defaults.txt'.
/Users/rmeli/Documents/git/software/UserGuide/doc/source/topology_system.rst:287: CRITICAL: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: 'source/generated/topology/groupmethods.txt'.
[...]
```

IAlibay: I'm going to say at this point, let's raise an issue and deal with it later? If @lilyminium says there's not much difference either way then it should be fine here for a bit.

IAlibay (approved):
Apologies @RMeli I didn't know I was the one with the blocking review here 😱 

lgtm!

Issue #205, opened 2022-04-27
Changes made in dependecy list for smooth installation


aditi2906:
There are a few suggested changes in the installation requirements of MDanalysis UserGuide to make it beginner-friendly.
After cloning the repo and creating a virtual environment:
a)Extracting dependency list from environment.yml to requirments.txt
b)Removing python dependency present in requirments.txt
c)resolving conflict between two different versions of jinja2 (accepting jinja2<3.1)
d)running pip install wheel
e)pip install markup 2.0.1

The above changes helped me install and run UserGuide smoothly on my local machine.

System: Ubuntu 20.04.4 LTS
installing using: pip

jandom:
Maybe we can close this one out? It's nearly a year old and there was a [merged PR](https://github.com/MDAnalysis/UserGuide/pull/229) that potentially fixed it?

PR #206, opened 2022-04-27
changes in requirments.txt for installation

aditi2906:
Solution code to requested feature #205


IAlibay:
Unfortunately @lilyminium who knows most about how things are pinned in the userguide is a little bit swamped with end-of-PhD things this month. I'll see if I can work out what exactly is required, but I'm not convinced that the solution here is to pin more things, instead we should be looking to relax our requirements a little bit.

aditi2906:
The requirements listed are already present in environment.yml and conda env create python=3.6 -f environment.yml --quiet 
is already using it to create virtual env. pip installing requirement.txt is failing the local build due to outdated requirements. So here I just updated requirements.txt to sync with environment.yml. If incase lesser dependencies are required in environment.yml, it would definitely relax the case.

IAlibay:
Closing as superseded by #229

PR #221, opened 2022-09-26
Update the UserGuide environment - make sure the latest MDA is working

IAlibay:


IAlibay:
aha! so it was because of the missing `__init__.py` entry... how silly.

In any case, this won't be backport-able for 2.3.0, so we might have to find a 2.3.0 specific solution (maybe just remove parmed from the userguide? that'd be a pain) and then deal with the rest later.

orbeckst:
Do the passing tests indicate that the UG is correctly building again?

I had a quick look and could give a superficial LGTM.

IAlibay:
> Do the passing tests indicate that the UG is correctly building again?
> 
> I had a quick look and could give a superficial LGTM. 

At the moment yes, however this is not backwards compatible with MDA 2.3.0, because of that missing init.py file. So what I need to do:

- hack in the parmed attributes
- merge into master to get 2.3.0 to deploy
- add the parmed dynamic reading in for 2.4.0-dev0
- look into moving converters into their own subsection eventually

IAlibay:
Note: there was some behind the scenes working out why RTD wasn't building. Turns out it was some background issue with conda updating itself, thankfully the RTD admin team had a flag they could use to disable this for the repo and things started to work again.

IAlibay:
@lilyminium it's a great solution, but it might be the best we can do at this moment in time given time constraints.

This temporarily removes ParmedParser from the table, avoiding the missing init file issue for 2.3.0, which should allow us to deploy 2.3.0 and move on (I'd prefer fix things for 2.4.0-dev0 than try to implement a hacky fix?).

If you're ok with this, it should be good to merge.

lilyminium (approved):
LGTM, thanks for this.

The automated scripting really needs a re-haul, tbh

PR #224, opened 2022-10-26
Remove hole.ipynb

IAlibay:


IAlibay:
Turns out it's pretty much the exact same notebook as hole2.ipynb, so there's really not reason to keep it around? (unless I'm missing something)

lilyminium (approved):
I believe one was for `hole2` the module (revamped) and one was the now-deprecated `hole.py` analysis. Thanks!

IAlibay:
whelp that was not meant to be targetting `master`

IAlibay:
doesn't really matter anyways since we're ditching `master`, I'll open a separate one for `develop`

PR #228, opened 2022-10-26
Fix mdanalysis install for `master` branch

IAlibay:
This should allow us to deploy for 2.3.0


IAlibay:
Self merging to move things along

PR #229, opened 2022-10-26
Fix environment.yaml and requirements.txt

IAlibay:


IAlibay:
@lilyminium The user guide states being able to install requirements via pip, but `pandoc` is an nbsphinx requirement. Am I correct in understanding that it's not really simple to install without conda? Would it make sense to just remove the requirements.txt file and switch over to instructions that are purely conda based?

lilyminium:
If you like, but pandoc and nbsphinx are both on pypi -- are you running into install issues?

IAlibay:
> If you like, but pandoc and nbsphinx are both on pypi -- are you running into install issues?

yup pandoc can't actually be installed via pypi, see: https://pypi.org/project/pandoc/

You essentially need to either install it manually or conda install it (it's a weird world where you have to do both).

IAlibay:
Ok I've gone ahead and updated `requirements.txt` and clarified the need to install `pandoc` is `pip` installing. Should be good for review.

lilyminium (approved):
LGTM, thanks!

PR #231, opened 2022-10-26
[wip] Can we use -j2 on make?

IAlibay:


IAlibay:
Absolutely no impact, closing.

PR #233, opened 2022-10-26
Fix hole notebook - this time for `develop`

IAlibay:


IAlibay:
Merging as previously approved -  the same as #224 just targetted at develop.

PR #253, opened 2023-04-09
Fix deploy on tag workflow

IAlibay:
I think some changes in the pip behavious made is such that the release install wouldn't override the dev0 version installed. So to go around this, we instead force a reinstall which should make it pick up whatever is on pypi.


IAlibay:
Once this is merged we need to pull the latest tag and redeploy so we at least have a 2.4.3 version out there.

orbeckst (approved):
I am just going to merge so that we can proceed from here.

PR #251, opened 2023-01-13
Fix nbval testing failures due to new 0.10 release

pgbarletta:
From #248 and #249

- I filed an issue on nbval(https://github.com/computationalmodelling/nbval/issues/194) where I explain what changed.
For now, I removed all semicolons that were used to silence cell outputs and added the necessary additional sanitize statements on `tests/sanitize_output.cfg`
- Also unpinned nbval

One small drawback is that for now I can't silence this ugly plots:
ugly_plot

PR #256, opened 2023-06-04
Remove widget states from notebooks

IAlibay:
I've gone ahead and cleared all the widget states from all the notebooks under `examples`. This reduces the total size of the `doc` files from 154 MB to 60 MB. 

As far as I'm aware this _shouldn't_ affect how the notebooks render, but someone please double check the RTD results.


pgbarletta:
I did the same when I added the nbval testing. They seem to accumulate with each run.

IAlibay:
> I did the same when I added the nbval testing. They seem to accumulate with each run.

they shouldn't be getting committed, I think it's more of a case that someone forgot to clear them after an update

lilyminium:
I *think* this removes the ability to look interactively at NGLView widgets on rendered docs. However I think the benefits of keeping the interactivity is much outweighed by the costs of storage compared to, say, a gif.

lilyminium:
And the huge diffs are a pain.

lilyminium (approved):
LGTM. Not sure how this will interact with #251 so I'll let you merge.

lilyminium:
Actually I'll merge now so I can build off it and replace widgets with gifs.

PR #267, opened 2023-07-14
Don't pull down cuda every time you do CI

IAlibay:
Small carbon emission savings.


lilyminium (approved):
Oh, that's what that shim does. Thank you!

PR #273, opened 2023-07-21
ci: enable micromamba env cache

jandom:
The micromamba environment creation takes about 5 minutes and is more than the actual build step (2 mins for buliding docs). Here is an example run

https://github.com/MDAnalysis/UserGuide/actions/runs/5622842883/job/15236245154

This is not ideal – developers want to see a crisp output of their work. 

I'm not sure how to enable this feature, or how well it works, so using this branch to experiment. If it works well, maybe we can consider porting to MDA core?

**Update** Here are some preliminary results showing the impact on the build speed, with and without caching

| Workflow Name | Caching on/off | Setup Duration | Job Duration |
|--------|--------|--------|--------|
| [mda_gh_ci](https://github.com/MDAnalysis/UserGuide/actions/runs/5623009172/attempts/1?pr=273) | Off | 4m 39s | 6m 16s |
| [mda_gh_ci](https://github.com/MDAnalysis/UserGuide/actions/runs/5623009172/attempts/2?pr=273) | On | 21s | 2m 6s |
| [mda_gh_ci_tests](https://github.com/MDAnalysis/UserGuide/actions/runs/5623009170/attempts/1?pr=273) | Off | 6m 29s | 16m 39s | 
| [mda_gh_ci_tests](https://github.com/MDAnalysis/UserGuide/actions/runs/5623009170/attempts/2?pr=273) | On | 27s | 15m 26s |

### DIFF
--- a/.github/workflows/gh-ci-tests.yaml
+++ b/.github/workflows/gh-ci-tests.yaml
@@ -35,7 +35,6 @@ jobs:
           environment-file: environment.yml
           environment-name: mda-user-guide
           cache-environment: true
-          cache-environment-key: gh-ci-tests
           create-args: >-
             hole2
 
--- a/.github/workflows/gh-ci-tests.yaml
+++ b/.github/workflows/gh-ci-tests.yaml
@@ -36,14 +36,6 @@ jobs:
           environment-name: mda-user-guide
           cache-environment: true
           cache-environment-key: gh-ci-tests
-          create-args: >- 
-            python==3.9
-            hole2
-            pip
-          condarc: |
-            channels:
-              - conda-forge
-              - plotly
 
       - name: "environment information"
         run: |
--- a/.github/workflows/gh-ci-tests.yaml
+++ b/.github/workflows/gh-ci-tests.yaml
@@ -35,6 +35,7 @@ jobs:
           environment-file: environment.yml
           environment-name: mda-user-guide
           cache-environment: true
+          cache-environment-key: gh-ci-tests
           create-args: >- 
             python==3.9
             hole2
--- a/.github/workflows/gh-ci-tests.yaml
+++ b/.github/workflows/gh-ci-tests.yaml
@@ -34,6 +34,7 @@ jobs:
         with:
           environment-file: environment.yml
           environment-name: mda-user-guide
+          cache-environment: true
           create-args: >- 
             python==3.9
             hole2
--- a/.github/workflows/gh-ci-tests.yaml
+++ b/.github/workflows/gh-ci-tests.yaml
@@ -36,6 +36,8 @@ jobs:
           environment-name: mda-user-guide
           cache-environment: true
           cache-environment-key: gh-ci-tests
+          create-args: >-
+            hole2
 
       - name: "environment information"
         run: |
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -46,13 +46,6 @@ jobs:
         environment-name: mda-user-guide
         cache-environment: true
         cache-environment-key: gh-ci
-        create-args: >- 
-          python=3.9
-          pip
-        condarc: |
-          channels:
-            - conda-forge
-            - plotly
 
     - name: install_deps
       run: |
--- a/.github/workflows/gh-ci.yaml
+++ b/.github/workflows/gh-ci.yaml
@@ -45,6 +45,7 @@ jobs:
         environment-file: environment.yml
         environment-name: mda-user-guide
         cache-environment: true
+        cache-environment-key: gh-ci
         create-args: >- 
           python=3.9
           pip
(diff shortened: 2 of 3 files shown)


IAlibay:
> maybe we can consider porting to MDA core?

We can't unfortunately. I already looked into this some time ago. Because of the size of the matrix the amount of data stored becomes excessive. Our build times are tiny so it's not a big deal tbh.

jandom:
That makes sense – yeah I can see how the build matrix can inflate the artifact size – but it's still a nice improvement to this repo?

.github/workflows/gh-ci-tests.yaml
```diff
@@ -34,6 +34,7 @@ jobs:
         with:
           environment-file: environment.yml
           environment-name: mda-user-guide
+          cache-environment: true
```
IAlibay, line 37: From a quick look through, it seems like you're getting a cache clash because both actions are using the same key: https://github.com/MDAnalysis/UserGuide/actions/runs/5623009170/job/15237260079#step:3:38

The safest solution here would be to just pass a different `cache-environment-key` for each action. That (edit: or) flatten the differences between the two workflows and put everything in the environment.yml.

IAlibay:
> That makes sense – yeah I can see how the build matrix can inflate the artifact size – but it's still a nice improvement to this repo?

I think so, but I'm trying to properly look at the impact, I might need to punt a complete look to the weekend.

1. There's a branch, develop precendence here that we need to consider. So the improvements only exist _if_ CI ran on `develop` earlier in the day, or if you ran CI in the PR earlier in the same day.
2. There's a cache clash I'm going to quickly write a review comment about now.
3. We can't change the cache time limit or enforce an override. So it looks as if your PR suddenly requires an upstream fix, you'll need to wait a day until the cache invalidates?

IAlibay (changes requested):
One quick initial comment.

IAlibay:
> There's a branch, develop precendence here that we need to consider. So the improvements only exist if CI ran on develop earlier in the day, or if you ran CI in the PR earlier in the same day.

From initial thoughts, I don't really care about this too much. At worse you get no improvement, at best it helps. It doesn't look like we'd get a regression so that's fine.

> We can't change the cache time limit or enforce an override. So it looks as if your PR suddenly requires an upstream fix, you'll need to wait a day until the cache invalidates?

This I care about a lot more. A lot of the userguide is "emergency patch fixes". I am concerned it might slow us down in those cases (unless I'm missing something that allows us to invalidate the cache easily).

jandom:
I've addressed the cache clash, so that should be fine – good catch!

Very wise re emergency invalidation but I think we can just wipe the caches here? https://github.com/MDAnalysis/UserGuide/actions/caches Optional tho, could be just more faff than it's worth – I'm 50/50 in my head right now

.github/workflows/gh-ci-tests.yaml
jandom: Actually, thinking about it, flattening will create a simpler situation, where we don't have to use the `cache-environment-key` feature

IAlibay: It might be good to do both even if they have the same key. That way we've got extra security if someone does changes in the future (it's one of those "someone new reads the entry it and realises how it all work" extra security steps).

IAlibay:
> I've addressed the cache clash, so that should be fine – good catch!
> 
> Very wise re emergency invalidation but I think we can just wipe the caches here? https://github.com/MDAnalysis/UserGuide/actions/caches Optional tho, could be just more faff than it's worth – I'm 50/50 in my head right now

Yeah (edit: deleting) cache manually might be enough. I don't really mind tbh, @MDAnalysis/coredevs does anyone have strong feelings here?

.github/workflows/gh-ci-tests.yaml
jandom: Actually @IAlibay I don't think there is a problem with a cache clash: yes the file hash is the same but the args has is different. The output below is from the two runs linked in the table above (where the cache was on), the `user-guide-args` is different


>  Restored cache with key `micromamba-environment--linux-64-mda-user-guide-args-958ca3f-file-69dbdc0bcba969b5795a9da29bcf23f024e7c0672d0bd1630afe58a4cb67706f`


>  Restored cache with key `micromamba-environment--linux-64-mda-user-guide-args-93344c2-file-69dbdc0bcba969b5795a9da29bcf23f024e7c0672d0bd1630afe58a4cb67706f`


I'd probably remove the cache key and keep this simpler, this github actions is clearly clever enough. Cherry on the cake we improved the environment.yml

RMeli:
No strong feelings, but I think that if the cache can be removed manually, then it's OK.

lilyminium (approved):
No strong feelings, as long as we remember the cache might need clearing then this seems an overall improvement. Thanks @jandom

IAlibay (approved):

jandom:
hi there many thanks for all the reviews and the merge (although that one was unexpected)

PR #274, opened 2023-07-22
Document darker usage

### LINKED ISSUE #247: Document linting with darker
Darker linting was added in [this PR](https://github.com/MDAnalysis/mdanalysis/pull/3954).

I think this is pretty much the information a developer needs:

Install:
```
conda install -c conda-forge darker flake8
````
Usage (assuming we forked from `develop` branch):
```
darker --check --diff --color -r upstream/develop -L flake8  package/MDAnalysis
darker --check --diff --color -r upstream/develop -L flake8  testsuite/MDAnalysisTests
```

### PR DESCRIPTION
**Pull request by pgbarletta**
Fixes #247

[This](https://userguide.mdanalysis.org/stable/contributing_code.html#adding-code-to-mda) is another section of the documentation where this issue is discussed, though it does recommend using `black` as a formatter, which kind of goes against the philosophy behind the usage of darker, since black doesn't allow partial reformatting.

Let me know if you want me to harmonize the 2 sections.

### CODE REVIEW

**Comment by IAlibay** [comment]
I think there's a misconception here, we never ask folks to pass the darker requirements, only the flake8 ones. The only reason we have the darker bot is because there are no other tools out there that run a minimal pep8 / flake8 check on diffs. (edit: what I mean here is that I do appreciate the optional label, but we need to make it clear that it's only flake/pep8 on diffs that matter)

**Comment by pgbarletta** [comment]
>  but we need to make it clear that it's only flake/pep8 on diffs that matter)

right, I thought this line was supposed to show them just that:
```
darker --diff -r upstream/develop package/MDAnalysis -L flake8
```
Or does flake8 enforce other checks beyond PEP8? 

If so, how do users know the difference between a PEP8 violation like [this one](https://github.com/MDAnalysis/mdanalysis/pull/4162#discussion_r1254985463):

image

and any other formatting requirement like breaking lists into 1-element-per-line?

**Comment by IAlibay** [comment]
I'm not on my desktop at the moment, but if I remember correctly, that line will show both `black` and `flake8` issues on the diff? I think what should be emphasized is that the only thing that absolutely needs addressing are the actual flake8 warnings at the bottom of the report, not the full output.

My point being that if they want to `black` format the whole diff that's fine, but if it conflicts with another linter they enjoy or readability then that's their call.

**Comment by pgbarletta** [comment]
Ok, now I get it. I think this should do.

**Review by IAlibay — COMMENTED** [review]
Thanks this looks reasonable to me from a very quick review, but I'll let another coredev do the proper review. I'm throwing this at @fiona-naughton but if someone else has time please chime in.

File: doc/source/contributing_code.rst [hunk 1]
```diff
@@ -473,6 +474,51 @@ Documenting your code
 
 Changes to the code should be reflected in the ongoing ``CHANGELOG``. Add an entry here to document your fix, enhancement, or change. In addition, add your name to the author list. If you are addressing an issue, make sure to include the issue number.
 
+.. _format-darker:
+
+-------------------------------------------------------------------------------
+Ensure PEP8 compliance (mandatory) and format your code with Darker (optional)
+-------------------------------------------------------------------------------
+
+`Darker <https://github.com/akaihola/darker>`_ is a *partial formating* tool that helps to reformat new or modified code
```
fiona-naughton:
```suggestion
`Darker <https://github.com/akaihola/darker>`_ is a *partial formatting* tool that helps to reformat new or modified code
```

fiona-naughton (commented):
just one little spelling fix, but otherwise lgtm!

fiona-naughton (approved):
Great - thanks @pgbarletta !

PR #276, opened 2023-07-24
feat: matrix builds for python versions

### LINKED ISSUE #166: Update testing
Right now testing is on 3.7 or 3.8 because I'm lazy, but it should really run on all the Python versions supported by MDA to check the notebooks.

jandom:
Maybe fixes https://github.com/MDAnalysis/UserGuide/issues/166

The general philosophy being: if you think your users can run it, why don't you confirm it yourself. Making assumptions about python versions has bitten me enough times, CPUs are cheap and people like libs that just work

### DIFF
--- a/.github/workflows/gh-ci-tests.yaml
+++ b/.github/workflows/gh-ci-tests.yaml
@@ -23,6 +23,9 @@ defaults:
 
 jobs:
   tests:
+    strategy:
+      matrix:
+        version: ["3.9", "3.10", "3.11"]
     runs-on: ubuntu-latest
     name: "examples-tests"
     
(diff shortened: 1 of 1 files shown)


IAlibay (changes requested):
I'm going to vote against this. The decision at the time was that this was already a super expensive test to run as-is so we decided to stick to the lowest Python version knowing that it'll be more or less always forwards compatible. We already run a matrix upstream to make sure that all the code here works.

jandom:
Ok, np - if there is no evidence that this leads to poor user outcomes, it's probably fine

jandom:
Hang on – if we don't want this, shouldn't we just close #166?

IAlibay:
I'm going to say yes but I'll leave it to the issue author to decide.

PR #275, opened 2023-07-24
feat: Add TNG to the list of readers

jandom:
Maybe fixes https://github.com/MDAnalysis/UserGuide/issues/243 - could it be so simple?

PR #278, opened 2023-07-25
Fix scripts

### LINKED ISSUE #277: Scripts aren't re-generating tables
**Describe the bug**

Scripts should be auto-generating format tables, including things like TNG. They should be failing right now since TNG isn't there. They aren't running.

lilyminium:
Fixes #277 

Due to #275 this should fail CI ~~and RTD~~


doc/source/scripts/generate_all.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e  # exit on failure
```
lilyminium, line 3: This is the most important part of the PR -- we exit the docs build if the scripts fail, which they will do if entries such as TNG are missing.

lilyminium:
This PR breaks CI on purpose -- if no one reviews or has comments, I'll merge in a few hours in anticipation of #275.

PR #280, opened 2023-07-29
refactor: harmonize bash scripts

jandom:
Make the bash scripts consistent in terms of error handling


lilyminium:
Thanks @jandom, these all look like good improvements and it's failing as it should. I think that `deploy_docs_via_travis.sh` is actually redundant and should be deleted now -- we moved on from Travis a while ago and at least in the user guide no longer use it for CI or deployment.

I'd hold off on approving/merging until #275 is done so we can check that it also succeeds when there's no error.

jandom:
Thanks @lilyminium , removed the travis deploy script now – thanks for the pre-review!

jandom:
@lilyminium PTAL this should be ready to go now, not blocked by anything

lilyminium (approved):
LGTM, thank you for this @jandom!

PR #286, opened 2023-08-05
dev: run black on the files

jandom:
run the black formatter on this codebase, only auto-generate code changes


maintainer/update_json_stubs_sitemap.py
```diff
@@ -23,29 +23,31 @@
...
-URL = os.environ['URL']
-VERSION = os.environ['VERSION']
+URL = os.environ["URL"]
+VERSION = os.environ["VERSION"]
 
 if "http" not in URL:
-    raise ValueError("URL should have the transfer protocol (HTTP/S). "
-                     f"Given: $URL={URL}")
+    raise ValueError(
+        "URL should have the transfer protocol (HTTP/S). " f"Given: $URL={URL}"
```
lilyminium:
```suggestion
        f"URL should have the transfer protocol (HTTP/S). Given: $URL={URL}"
```

maintainer/update_json_stubs_sitemap.py
```diff
@@ -23,29 +23,31 @@
...
+    )
 
 try:
     int(VERSION[0])
 except ValueError:
-    raise ValueError("$VERSION should start with a number. "
-                     f"Given: $VERSION={VERSION}") from None
+    raise ValueError(
+        "$VERSION should start with a number. " f"Given: $VERSION={VERSION}"
```
lilyminium:
```suggestion
        f"$VERSION should start with a number. Given: $VERSION={VERSION}"
```

lilyminium (approved):
Just a couple minor suggestions where combining strings into a single line was kind of weird; otherwise LGTM, thanks @jandom!

RMeli (approved):
As discussed in https://github.com/MDAnalysis/mdanalysis/issues/2450, `black` is (was?) still somewhat controversial. Since here the changes are applied only to internal scripts, I think this PR would not be too controversial. 

I'm happy to see  approval from @lilyminium on this for the UserGuide, so unless someone else objects I think we can move forward. 

(Maybe applying black to the Jupyter Notebooks, which are user-facing, will be more controversial, but we can discuss if/when that's the case.)

doc/source/conf.py
```diff
@@ -217,4 +226,6 @@ def sort_authors(filename):
 # rst-epilog implements substitutions
 rst_epilog = """
 .. |MDAnalysis_version| replace:: {0}
-""".format(MDAnalysis_version)
+""".format(
+    MDAnalysis_version
+)
```
RMeli: This is one of the weird ones, but harmless. And would be a non-issue with f-strings.

Maybe a first good step to apply black to the main code base would be to convert `format` statements into f-strings.

jandom: Good shout, I turned this into an f-string

PR #287, opened 2023-08-05
dev: add black to pre commit

jandom:
Adding black to pre-commit-config

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,3 +11,8 @@ repos:
     hooks:
       - id: isort
         name: isort (python)
+        args: ["--profile", "black"]
+-   repo: https://github.com/psf/black
+    rev: 22.3.0
+    hooks:
+    -   id: black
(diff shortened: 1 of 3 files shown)


lilyminium (approved):
LGTM, thanks @jandom.

You have quite a few PRs approved and ready to merge now, if you want to handle them! We typically squash-merge to avoid lots of small commits with minor changes (not relevant to this PR but some others are a bit larger).

RMeli (changes requested):
LGTM, thanks @jandom! It just need the commit of #286 (once squash-merged) in the `.git-blame-ignore-revs` and then it's good to go!

.pre-commit-config.yaml
```diff
@@ -11,3 +11,8 @@ repos:
     hooks:
       - id: isort
         name: isort (python)
+        args: ["--profile", "black"]
```
jandom, line 14: forgot this about curve-ball: isort and black can be in conflict without this

jandom:
@RMeli this should be cleaned-up now, PTAL

.pre-commit-config.yaml
RMeli: Good catch!

PR #289, opened 2023-08-06
refactor: remove dead code

jandom:
snapshot tests in the [other PR](https://github.com/MDAnalysis/UserGuide/pull/285) identified dead code that's never executed, we can safely removed it. The snapshot tests ran all the script code, so there is no risk of missing a code-path. Removing this code will make the code simpler to reason about, and make refactoring easier.

### DIFF
--- a/doc/source/scripts/base.py
+++ b/doc/source/scripts/base.py
@@ -25,11 +25,8 @@ class TableWriter(object):
     postprocess = []
     sort = True
 
-    def __getattr__(self, key):
-        try:
-            return self.fields[key]
-        except KeyError:
-            return super(TableWriter, self).__getattr__(key)
+    def __getattr__(self, key: str) -> list:
+        return self.fields[key]
 
     def __init__(self, *args, **kwargs):
         stem = os.getcwd().split("source")[0]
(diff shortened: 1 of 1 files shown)


doc/source/scripts/base.py
```diff
@@ -25,11 +25,8 @@ class TableWriter(object):
...
-    def __getattr__(self, key):
-        try:
-            return self.fields[key]
-        except KeyError:
-            return super(TableWriter, self).__getattr__(key)
+    def __getattr__(self, key: str) -> list:
+        return self.fields[key]
 
     def __init__(self, *args, **kwargs):
         stem = os.getcwd().split("source")[0]
```
lilyminium, line 32: (Non-blocking) this has always bothered me, I think starting from `__file__` is probably more robust, and using the `pathlib` library too -- but out of scope for this PR.

doc/source/scripts/base.py
```diff
@@ -99,9 +96,7 @@ def sphinx_meth(meth, tilde=True):
         return ":meth:`{}{}.{}`".format(prefix, meth.__module__, meth.__qualname__)
 
     @staticmethod
-    def sphinx_ref(txt, label=None, suffix=""):
-        if label is None:
-            label = txt
+    def sphinx_ref(txt: str, label: str = None, suffix: str = "") -> str:
         return ":ref:`{} <{}{}>`".format(txt, label, suffix)
```
lilyminium: As Rocco said in another PR, it'd be great to move solely onto f-strings; I think the user guide was started when we still supported 2.7, so there's a lot of now unnecessary contortions in the code. Out of scope here though!

lilyminium (approved):
LGTM -- very nice to see modern typing here! I believe `list` is 3.9 onwards and that we support the most recent 3 versions of Python, so we should be good here -- I'm actually not sure we say anywhere explicitly which Python versions is supported by the user guide.

doc/source/scripts/base.py
jandom: Let's do it here, it'd be simple enough

jandom:
Good shout, yeah I switched to using f-strings here and rebased onto master. Will merge after the tests pass

PR #292, opened 2023-08-07
dev: remove requirements.txt

jandom:
During the discussion in https://github.com/MDAnalysis/UserGuide/pull/290 we've aligned (I think...) on dropping the pip workflow. It was broken anyway and never exercised in CI/CD. Sticking with just conda-like environments is probably the simplest thing for now, and lower the maintenance surface area for the project. 

Next steps maybe some sort of dependency cleanup, with better isolation of pip packages, and maybe a pyproject.toml (out of scope here)


lilyminium (approved):
LGTM -- thanks for removing the old python pin!

jandom:
Thanks no worries – this one is a slam dunk, things just getting simpler

PR #293, opened 2023-08-07
dev: cleanup environment.yml

jandom:
Just an experiment, I tried to make sense of the environment.yml – there is a LOT there. I was able to run locally

```
make -C doc html
```

But I want to see if the tests pass in CI/CD


environment.yml
```diff
@@ -2,60 +2,74 @@ name: guide
...
+      - ipywidgets<8.0.0
+      - nbsphinx
+      # used for notebook testing
       - nbval
+      # building documentation
+      - tabulate
+      # tests for this package
       - pre-commit
-      - pyedr>=0.7
+      - pytest
```
IAlibay, line 72: By moving a lot of stuff out of `conda` into `pip` installs you're effectively just saying "nah I don't feel like good dependency resolution". It's in my opinion significantly worsening the fragility of the environment build?

jandom: It's not just that but yeah I see your point. Just wanted to push things around, it's an exploratory PR. It ideally shouldn't matter if the package is from pip vs conda channel. 

For a user guide package, there are many deps and many were found to be obsolete (the build still passes). From things under `pip` we can more easily migrate to subpackages in pyproject.toml (where people who just want to build don't need to install syrup, or some other tests framework).

jandom: It's just quite hard for a new comer to understand "why all these deps? what is each needed for?"

The more we can put into pip, the more understandable things will get in my view – it'll surface things that are "conda-only" and that will be communicated explicitly. `openmm` for example is only available in conda, same for graphviz, and a few others. it's not clear that's the case until you start messing about ;-)

The last cherry is that the env is now much slimmer: 1.8 GB vs 2.6 GB. But that's a bonus objective, just improving on the margins.

lilyminium: I'm not sure the benefit of communicating what is only on conda vs what is only on pip is worth the additional [environment fragility](https://www.anaconda.com/blog/using-pip-in-a-conda-environment). Perhaps I'm not fully understanding, since you've also left a number of packages in the conda part (e.g. matplotlib) that are available on pypi. I think we could turn this around and communicate "these packages are *not* available on conda-forge" with the `pip` section, which as I understand it is the standard way to structure these environment files.

> It's just quite hard for a new comer to understand "why all these deps? what is each needed for?"

I think the comments you've put in separating the sections into mda/docs sections are better at communicating this, and go a long way to making it much clearer!

> The last cherry is that the env is now much slimmer: 1.8 GB vs 2.6 GB. But that's a bonus objective, just improving on the margins.

I assume much of this is eliminating CUDA and other unnecessary packages?

environment.yml
```diff
@@ -1,61 +1,76 @@
...
   - pandoc
-  - moviepy
-  - ipyvtklink
+  - pybtex
+  - sphinx
+  - sphinxcontrib-bibtex
+  # - jedi # not used?
+  # - jinja2 # not used?
+  # - nbclient # not used?
+  # - parso # not used?
```
lilyminium, line 29: You're probably good to remove these. For some history, at one point the user guide environment was having trouble building in good time on RTD, so we pinned down some dependencies to help resolution. Also, some of these (e.g. jinja2, which is used under the hood in sphinx and other tempaltes) had to be pinned to a specific version in the past for issues that have been since resolved.

environment.yml
jandom: That's a good point @lilyminium I wasn't super consistent moving everything under pip. 

Okay, let's take a step back here – I concede that this PR is trying to do too much – I'll undo the these pip/conda reshuffles, keep the comments and ditch the packages we no longer need

environment.yml
jandom: What's RDT? Makes sense re jinja2 and other such "creatures" I suspect there are a few more obsolete packages. The systematic way to eliminate these is to create a single PR for each dep, and confirm if the build works or not... but I don't recommend doing that, people get grumpy :D

lilyminium: RTD stands for ReadTheDocs -- it's one of the checks we run for PRs. It now uses mamba, but when the user guide was first started, our builds would occasionally time out waiting for conda to resolve :)

jandom: Ah, gotcha read the docs. Yeah the resolution time with native conda is just bafflingly slow... turning environments into lock files really helps, have you ever seen that tried in the MDA-verse?

IAlibay: It's been discussed in the past (for the core library) but the general take is that lock files aren't really appropriate for libraries.

My personal take is that they have a space for when you need ironclad production releases but in places like CI create a "leave it and forget it" landmine for some future maintainer to deal with. Indeed this is sort of what happened when we pinned nearly every dependency in this repo.

In any case `mamba` has mostly resolved our issues now.

PR #295, opened 2023-08-08
dev: comment and cleanup environment.yml

jandom:
Lots of discussion in https://github.com/MDAnalysis/UserGuide/pull/293 big thanks to @IAlibay and @lilyminium for helping me out. 

This PR removes a few packages and adds a lot of comments. It's still possible that there are other unused packages.

The aim of this PR is to make it clearer which dep is needed where, it's still not super granular, and it's still possible that other packages are not needed.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -9,7 +9,6 @@ dependencies:
   - openmm>=7.0.0
   - parmed
   - scikit-learn>=1.1.2
-  - threadpoolctl
   - pyedr>=0.7
   # building documentation
   - docutils
--- a/environment.yml
+++ b/environment.yml
@@ -22,9 +22,12 @@ dependencies:
   - sphinxcontrib-bibtex
   - tabulate
   # extensions
+  - ipython
+  - ipywidgets<8.0.0
   - jupyter_contrib_nbextensions
   - nbconvert # not called by any notebook explicitly but to download notebooks, or in nbsphinx
   - nbformat
+  - nbsphinx
   - notebook<7
   # used for visualising/plotting
   - graphviz
@@ -40,10 +43,6 @@ dependencies:
   # used for visualising/making movies of molecules;
   - nglview
   - moviepy
-  # used in conf.py
-  - ipython
-  - ipywidgets<8.0.0
-  - nbsphinx
   # used for notebook testing
   - pytest
   - python=3.9
--- a/environment.yml
+++ b/environment.yml
@@ -7,7 +7,6 @@ channels:
 dependencies:
   # mdanalysis dependencies
   - openmm>=7.0.0
-  - packaging
   - parmed
   - scikit-learn>=1.1.2
   - threadpoolctl
--- a/environment.yml
+++ b/environment.yml
@@ -5,57 +5,66 @@ channels:
   - plotly
   - defaults
 dependencies:
+  # mdanalysis dependencies
   - cython
-  - distributed==2021.1.0
-  - dask==2021.1.0
-  - python-graphviz
-  - pytest
-  - decorator
-  - docutils
-  - graphviz
-  - ipywidgets<8.0.0
-  - ipython
-  - ipygany
-  - jedi
-  - jinja2
-  - jupyter_contrib_nbextensions
   - matplotlib
-  - nbclient
-  - nbconvert
-  - nbformat
-  - nbsphinx
   - networkx
-  - nglview
-  - notebook
   - numpy
+  - openmm>=7.0.0
   - packaging
   - parmed
-  - parso
-  - pip
-  - plotly
-  - pybtex
-  - python=3.9
-  - pyvista
-  - scikit-image
   - scikit-learn>=1.1.2
+  - threadpoolctl
+  # building documentation
+  - docutils
+  - pandoc
+  - pybtex
   - sphinx
   - sphinxcontrib-bibtex
-  - threadpoolctl
+  # used by custom_parallel_analysis.ipynb
+  - graphviz
   - tabulate
-  - widgetsnbextension
-  - ipycytoscape>=1.3.0
+  - python-graphviz
+  # used in multiple notebooks
+  - jupyter_contrib_nbextensions
+  - nbconvert
+  - nbformat
+  - nglview
+  - notebook<7
   - seaborn>=0.12.1
-  - openmm>=7.0.0
-  - pandoc
+  # used in pairwise_rmsd.ipynb
+  - plotly
+  # used in custom_parallel_analysis.ipynb
+  - distributed==2021.1.0
+  - dask==2021.1.0
+  # used in density_analysis.ipynb
+  - ipygany
+  # used in density_analysis.ipynb
+  - pyvista
+  # used in density_analysis.ipynb
+  - scikit-image
+  # used in aligning_trajectory.ipynb
   - moviepy
-  - ipyvtklink
+  # used in conf.py
+  - ipython
+  - ipywidgets<8.0.0
+  - nbsphinx
+  # used for notebook testing
+  - pytest
+  - python=3.9
+  - pip
   - pip:
-      - sphinx_rtd_theme
+      # used in conf.py
       - msmb_theme
+      - sphinx_rtd_theme
       - sphinx-sitemap
+      # used for notebook testing
       - nbval
+      # developer tooling
       - pre-commit
+      # mdanalysis dependencies
       - pyedr>=0.7
+      # mdanalysis and its tests and test data
       - git+https://github.com/MDAnalysis/mdanalysis@develop#egg=mdanalysis&subdirectory=package
       - git+https://github.com/MDAnalysis/mdanalysis@develop#egg=MDAnalysisTests&subdirectory=testsuite
       - git+https://github.com/MDAnalysis/MDAnalysisData@master#egg=MDAnalysisData
--- a/environment.yml
+++ b/environment.yml
@@ -6,15 +6,14 @@ channels:
   - defaults
 dependencies:
   # mdanalysis dependencies
-  - cython
   - matplotlib
   - networkx
-  - numpy
   - openmm>=7.0.0
   - packaging
   - parmed
   - scikit-learn>=1.1.2
   - threadpoolctl
+  - pyedr>=0.7
   # building documentation
   - docutils
   - pandoc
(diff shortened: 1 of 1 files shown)


environment.yml
```diff
@@ -5,57 +5,66 @@ channels:
...
+  # used in density_analysis.ipynb
+  - ipygany
+  # used in density_analysis.ipynb
+  - pyvista
+  # used in density_analysis.ipynb
+  - scikit-image
+  # used in aligning_trajectory.ipynb
   - moviepy
-  - ipyvtklink
+  # used in conf.py
```
lilyminium: I mostly think of these as sphinx "extensions" -- we use the `ipython` directive in some docs, `nbsphinx` to render notebooks, etc. If they're imported/used in `conf.py` it would generally be in their role as extensions.

environment.yml
```diff
@@ -5,57 +5,66 @@ channels:
...
   - sphinx
   - sphinxcontrib-bibtex
-  - threadpoolctl
+  # used by custom_parallel_analysis.ipynb
+  - graphviz
   - tabulate
-  - widgetsnbextension
-  - ipycytoscape>=1.3.0
+  - python-graphviz
+  # used in multiple notebooks
```
lilyminium: I would label this section generally `extensions` instead. `nbconvert` especially I would assume is not called by any notebook explicitly but to download notebooks, or in nbsphinx.

environment.yml
```diff
@@ -5,57 +5,66 @@ channels:
...
+  - jupyter_contrib_nbextensions
+  - nbconvert
+  - nbformat
+  - nglview
+  - notebook<7
   - seaborn>=0.12.1
-  - openmm>=7.0.0
-  - pandoc
+  # used in pairwise_rmsd.ipynb
+  - plotly
```
lilyminium: As we edit or update notebooks, I can see the notebook-specific labels getting outdated quite easily. Instead, could we maybe organise these by purpose? e.g. `plotly`, `ipygany`, `pyvista`, `scikit-image`, `seaborn` are for visualising/plotting; `moviepy` and `nglview` are for visualising/making movies of molecules; `dask` and `distributed` you could switch the heading to `# parallelization`.

lilyminium (changes requested):
Thanks @jandom for reorganising this file! I had some comments on the conceptual grouping. I'm not sure organising by notebook file is easiest to understand and maintain, especially since there's nothing stopping us from using e.g. `scikit-image` in another notebook and forgetting to update this. Instead, I've suggested ways to document by overall purpose.

environment.yml
```diff
@@ -5,57 +5,66 @@ channels:
   - plotly
   - defaults
 dependencies:
+  # mdanalysis dependencies
   - cython
```
IAlibay: It's odd that we only cover a small subset of the MDA setup and install requirements:

setup requirements: [Cython, packaging, numpy>=1.22.3, setuptools, wheel]
additional install requirements: [biopython>=1.8, networkx>=2.0, GridDataFormats>=0.4.0, mmtf-python>=1.0.0, joblib>=0.12, matplotlib>=1.5.1, tqdm>=4.43.0, threadpoolctl, fasterners]

environment.yml
```diff
@@ -5,57 +5,66 @@ channels:
...
+      # used in conf.py
       - msmb_theme
+      - sphinx_rtd_theme
       - sphinx-sitemap
+      # used for notebook testing
       - nbval
+      # developer tooling
       - pre-commit
+      # mdanalysis dependencies
       - pyedr>=0.7
```
IAlibay: this is available on conda-forge now: https://github.com/conda-forge/pyedr-feedstock

IAlibay (changes requested):

environment.yml
lilyminium: These might be covered by the `pip` install of `develop` MDA -- iirc only cython and numpy are crucial to have before installing MDA. At least CI is correctly building from this environment. I'd lean towards having as few dependencies here as possible so if core MDA updates its dependencies, it's reflected here as well.

IAlibay: Even cython and numpy aren't crucial. They get picked up on wheel creation (the magic of our pyproject.toml). I'd advocate for getting rid of all the non optional dependencies then (although it does add to the risk of a dependency resolution failure).

environment.yml
jandom: Thank you these changes were easy to apply

environment.yml
jandom: Thanks, yeah, easy to change

environment.yml
jandom: Changed the grouping here as well, easy to apply

environment.yml
jandom: i'm assuming you wanted this changed to a conda package? or was it just a "fyi"?

environment.yml
jandom: I'm not really sure what's the request here? I've removed numpy and cython from the environment.yml, i'm happy to proceed and remove other deps assuming the CI/CD passes – what's needed here?

environment.yml
```diff
@@ -5,57 +5,60 @@ channels:
...
+  - pybtex
   - sphinx
   - sphinxcontrib-bibtex
-  - threadpoolctl
+  # used by custom_parallel_analysis.ipynb
+  - graphviz
   - tabulate
-  - widgetsnbextension
-  - ipycytoscape>=1.3.0
+  - python-graphviz
```
lilyminium: Could you move this down to #visualisation?

environment.yml
lilyminium: Thanks @jandom -- could you please move these up to the #extensions then?

environment.yml
```diff
@@ -5,57 +5,60 @@ channels:
...
+  - ipython
+  - ipywidgets<8.0.0
+  - nbsphinx
+  # used for notebook testing
+  - pytest
+  - python=3.9
+  - pip
   - pip:
-      - sphinx_rtd_theme
+      # used in conf.py
```
lilyminium:
```suggestion
      # docs themes
```

environment.yml
```diff
@@ -5,57 +5,60 @@ channels:
...
       - msmb_theme
+      - sphinx_rtd_theme
       - sphinx-sitemap
```
lilyminium, line 51:
```suggestion
      # extensions
      - sphinx-sitemap
```

lilyminium (changes requested):
Hi @jandom, thank you for making the changes! Just a couple things would be nice to re-arrange and then I think we're good to go.

environment.yml
lilyminium: Personally I'd be happy with as minimal MDA requirements as possible here.

jandom: Agreed but it'll take a lot of iteration to remove everything but what's absolutely needed

environment.yml
lilyminium: Just this to go -- nbsphinx, ipywidgets and the like are used to render notebooks, and IIRC ipython to use the `ipython` sphinx directive. If they're imported in conf.py it's to configure settings but, otherwise they fall into the same category as other extensions, e.g. sphinx bibtex.

lilyminium (changes requested):

lilyminium:
I see @IAlibay is sick -- did you still want to re-visit this PR, Irfan? If not I can take over if you dismiss your review or approve changes.

IAlibay:
As far as I am aware (edit: from looking at my last request changes) there are still four build or core dependencies of
MDAnalysis left in the yaml file (matplotlib, networkx, packaging and
threadpoolctl). If those are included intentionally then it would be great
to document why, if not it would be good to remove them here.

If this PR needs a quick resolution then please go ahead and dismiss my
review.

jandom:
@lilyminium this should be resolved now, re extensions. I still have it in me to try and remove all the other packages that are MDA deps

environment.yml
```diff
@@ -5,57 +5,59 @@ channels:
...
-  - docutils
-  - graphviz
-  - ipywidgets<8.0.0
-  - ipython
-  - ipygany
-  - jedi
-  - jinja2
-  - jupyter_contrib_nbextensions
+  # mdanalysis dependencies
   - matplotlib
```
IAlibay:
```suggestion
```

environment.yml
```diff
@@ -5,57 +5,59 @@ channels:
...
-  - nbclient
-  - nbconvert
-  - nbformat
-  - nbsphinx
   - networkx
```
IAlibay:
```suggestion
```

environment.yml
```diff
@@ -5,57 +5,59 @@ channels:
...
-  - nglview
-  - notebook
-  - numpy
+  - openmm>=7.0.0
   - packaging
```
IAlibay:
```suggestion
```

environment.yml
```diff
@@ -5,57 +5,59 @@ channels:
...
   - parmed
-  - parso
-  - pip
-  - plotly
-  - pybtex
-  - python=3.9
-  - pyvista
-  - scikit-image
   - scikit-learn>=1.1.2
+  - threadpoolctl
```
IAlibay:
```suggestion
```

IAlibay (changes requested):

environment.yml
jandom: Let's try these all, thanks!

jandom:
Sorry a history question... please ignore if it's a time waste... but i'm really curious how all these MDA deps made their way into here?

IAlibay (approved):

jandom:
@IAlibay thanks for the confidence and just nuking those deps – this helped tremendously, I would have been removing them one-by-one, because I don't have the context

jandom:
@lilyminium PTAL, this seems to be blocked on an approval from you – and i don't think it merits an override (it's nor urgent)

IAlibay:
> Sorry a history question... please ignore if it's a time waste... but i'm really curious how all these MDA deps made their way into here?

They've been around since before we started supporting PEP518, i.e. a world before you could just rely on pip to pick up the right build time dependencies & before we offered wheels. Back then it was safer to have your own deps before you attempted to pip install something.

lilyminium (approved):
Sorry for the wait -- thanks for the changes @jandom and for all your patience!

jandom:
Thank you for all you reviews 👏

jandom:
Looking back at this month, we got a lot done – my original aim was to wrap up the snapshot tests and [refactor PR](https://github.com/MDAnalysis/UserGuide/pull/301), and I think we can do that too. @IAlibay and @lilyminium this wouldn't be possible without your engagement and encouragement, thank you!

PR #302, opened 2023-08-21
Add troubleshooting info to contribution docs

ALescoulie:
The last few times I've tried setting up an MDAnalysis development environment on Linux I've had compiler errors when using pip to build the package due to not having build-essentials installed, I figured I'd make a quick edit so others have an easier time troubleshooting.

### DIFF
--- a/doc/source/contributing_code.rst
+++ b/doc/source/contributing_code.rst
@@ -262,7 +262,7 @@ Ensure that you have a working C/C++ compiler (e.g. gcc or clang). You will also
         python -m pip install -e .
 
 
-.. note:: If installing  fails due to compiler ensure that you have the installed the build essentials package. It can be installed with your OS/distro's package manager. 
+.. note:: If installing  fails due to compiler ensure that you have the installed the build-essentials package. It can be installed with your OS/distro's package manager. 
 
 At this point you should be able to import MDAnalysis from your locally built version. If you are running the development version, this is visible from the version number ending in :code:`-dev0`. For example:
 
--- a/doc/source/contributing_code.rst
+++ b/doc/source/contributing_code.rst
@@ -261,6 +261,9 @@ Ensure that you have a working C/C++ compiler (e.g. gcc or clang). You will also
         cd ../testsuite/
         python -m pip install -e .
 
+
+.. note:: If installing  fails due to compiler ensure that you have the installed the build essentials package. It can be installed with your OS/distro's package manager. 
+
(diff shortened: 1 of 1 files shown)


IAlibay:
That's odd, you really shouldn't need to use build essentials. How are you building MDAnalysis? (these types of problems should have gone away with `pip install` builds).

edit: I guess this would happen anyways if you don't have gcc / g++ installed.

doc/source/contributing_code.rst
```diff
@@ -261,6 +261,9 @@ Ensure that you have a working C/C++ compiler (e.g. gcc or clang). You will also
         cd ../testsuite/
         python -m pip install -e .
 
+
+.. note:: If installing  fails due to compiler ensure that you have the installed the build-essentials package. It can be installed with your OS/distro's package manager.
```
IAlibay, line 265: Would this just fall under "ensure you have a working C/C++ compiler" above? (I assume that'ss what's missing here?)

Maybe we should expand that section and add the details of where you might find compilers on various OS there?

ALescoulie:
I was following the instruction on the contributing page and it failed with compiler errors at `pip install -e .` using a fedora dev container that didn't come with most default packages, I think I mostly needed the compilers installed and those are packaged with build-essentials, the specific error was `cc1plus`. I fixed it by installing g++. 

That might have just been that when I installed gcc I didn't get the C++ stuff.

doc/source/contributing_code.rst
ALescoulie: I think it might, and I just ended up installing build essentials to fix it

lilyminium: Is build-essentials specific to linux? I can't remember how I get things going on mac, but I can usually `brew install gcc` etc (or it might come as a click-installable developer tools as part of xcode)

IAlibay: It's specific to Debian flavours of Linux (so even more "obscure"). Unless we want to cover all major OS types and flavours I think the best that could be done is to just augment the C/C++ comment by saying you can get it from your favourite package manager?

lilyminium: I agree with @IAlibay here -- @ALescoulie would you be ok making that change?

RMeli:
> That might have just been that when I installed gcc I didn't get the C++ stuff.

Closing since it seems to be an issue with a system/compiler installation issue. @ALescoulie please feel free to re-open, if needed.

PR #304, opened 2023-08-30
Pin docutils for now

IAlibay:


IAlibay:
We're getting doc build failures because due to version mismatches.

IAlibay:
cycling - there's an issue with mamba it seems

lilyminium (approved):
LGTM -- hopefully we can remove the pin soon!

PR #305, opened 2023-08-30
Add RTD preview action

IAlibay:
Whilst I'm here and noticing that it's pretty hard to get to the RTD build...


lilyminium (approved):
Doesn't it come up as a check?
LGTM though -- thank you!

IAlibay:
It's a pull request target so it needs merging before it'll try to do anything.

PR #310, opened 2023-09-05
Update python setup.py build_sphinx

### LINKED ISSUE #308: Move on from build_sphinx
The docs say to use build_sphinx -- this is deprecated from sphinx 7.0 onwards. Instead we should use whatever solution is to https://github.com/MDAnalysis/mdanalysis/issues/4277

lilyminium:
Fixes #308 

Chooses `make html` over `build_sphinx`, which will be unsupported as of sphinx 7.0. `make html` is much simpler than having to specify the source and destination directories of `sphinx-build`. Since this is just building docs for personal use, I went with the simpler command.

----
:books: Documentation preview :books:: https://mdanalysis--310.org.readthedocs.build/en/310/

### DIFF
--- a/doc/source/contributing_code.rst
+++ b/doc/source/contributing_code.rst
@@ -687,7 +687,7 @@ Then, generate the docs with:
 
     .. code-block:: bash
 
-        python setup.py build_sphinx -E
+        cd doc/sphinx && make html
 
 This generates and updates the files in ``doc/html``. If the above command fails with an ``ImportError``, run
 


doc/source/contributing_code.rst
```diff
@@ -687,7 +687,7 @@ Then, generate the docs with:
 
     .. code-block:: bash
 
-        python setup.py build_sphinx -E
+        cd doc/sphinx && make html
```
IAlibay, line 690: So may be a controversial opinion, but if we're going to eventually encourage folks to run doctest on the regular, would it make sense here to use what we do in CI and call `sphinx-build`?

Maybe I don't understand the advantage of calling `make html` over `sphinx-build` though.

lilyminium: I find it simpler than `sphinx-build` since it reads the source and destination files from the Makefile, and it's an [endorsed way](https://www.sphinx-doc.org/en/master/usage/quickstart.html#running-the-build) to build docs. I'm not too attached either way.

IAlibay: Can we run doctests through the make call? If not, I'd prefer stick to sphinx-build.

lilyminium: Yup, `make doctest` -- it's in the Makefile.

IAlibay (approved):
Thanks for answering my questions, lgtm!

PR #313, opened 2023-09-14
Document TPRParser sets chainID topology attr

### LINKED ISSUE #312: document TPRParser now setting chainID
With @pgbarletta's  PR https://github.com/MDAnalysis/mdanalysis/pull/4281 merged (for issue https://github.com/MDAnalysis/mdanalysis/issues/3994), the TPRParser is now setting chainID for the topology.

This should be documented in the UserGuide, too. For example:
- https://userguide.mdanalysis.org/stable/formats/reference/tpr.html
- https://userguide.mdanalysis.org/stable/topology_system.html#format-specific-attributes
- ... find other pages

Long form description

> If the TPR molblock is named "Protein_chain_XXX" then we assume that XXX is   describing the chain of a protein (in the sense of the PDB chainID) and set   the topology attribute `chainID` to "XXX". In all other cases, the chainID   remains the full molblock name. The `segID` is never changed.

pgbarletta:
Fixes #312.

Added `chainID` to `tpr.rst`, which impacts on:
 * https://userguide.mdanalysis.org/stable/formats/reference/tpr.html
 
And ran `scripts/gen_topologyparser_attrs.py` which changed a lot of things, but I just committed the line for the TPR parser and the one above (sorry, git wouldn't let me split the hunk of the 2 lines into one hunk for each line).
This impacted on these files:

 * https://userguide.mdanalysis.org/stable/topology_system.html#format-specific-attributes
 * https://userguide.mdanalysis.org/stable/formats/index.html
 * https://userguide.mdanalysis.org/stable/formats/topology.html

----
:books: Documentation preview :books:: https://mdanalysis--313.org.readthedocs.build/en/313/

### DIFF
--- a/doc/source/formats/reference/tpr.rst
+++ b/doc/source/formats/reference/tpr.rst
@@ -58,14 +58,37 @@ TPR specification
 =================
 
 The TPR reader is a pure-python implementation of a basic TPR
-parser. Currently the following sections of the topology are parsed:
+parser. Currently the following topology attributes are parsed:
 
-* Atoms: number, name, type, resname, resid, segid, mass, charge,
+* Atoms: number, name, type, resname, resid, segid, chainID, mass, charge,
   [residue, segment, radius, bfactor, resnum, moltype]
 * Bonds
 * Angles
 * Dihedrals
 * Impropers
+
+``segid`` and ``chainID``
+-------------------------
+
+MDAnalysis gets the ``segment`` and ``chainID`` attributes from the TPR ``molblock`` field.
+Since TPR files are built from GROMACS topology files, ``molblock`` fields get their names from the
+compounds listed under the ``[ molecules ]`` header. For example::
+
+    [ molecules ]
+    ; Compound        #mols
+    Protein_chain_A     1
+    Protein_chain_B     1
+    SOL             40210
+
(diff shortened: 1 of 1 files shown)


lilyminium (changes requested):
Thanks @pgbarletta!

I'm going to focus on tpr.rst since I think we should actually not be committing txt files -- I think that's supposed to be generated on the fly when docs are built, and it was my mistake that they're currently in the repo. I'll investigate this separately -- just raised #314 to deal with it.

Could you please add a note somewhere explaining how the chainID is set, since there's some logic if it's a protein chain? Something like the note you left in the PR would be great, or at least a link to it in the docs:

```
   If the TPR molblock is named "Protein_chain_XXX" then we assume that XXX is 
   describing the chain of a protein (in the sense of the PDB chainID) and set
   the topology attribute `chainID` to "XXX". In all other cases, the chainID
   remains the full molblock name. The `segID` is never changed.
```

I know it's repetitive but ideally for behaviour that's not straightforward it gets documented in the user guide.

pgbarletta:
As usual, I ended up overextending myself, but I think it should be ok, given that we're on the UG.

lilyminium (approved):
This is a fantastic and super thorough explanation -- thanks so much @pgbarletta!

PR #316, opened 2023-09-15
Pin theme higher

lilyminium:
Develop docs are still pulling mda sphinx theme 1.0.1 -- pinning higher to force updates.

----
:books: Documentation preview :books:: https://mdanalysis--316.org.readthedocs.build/en/316/


IAlibay (approved):
Odd that they don't default to the newest version.

PR #345, opened 2023-12-01
Rerun diffusion_map.ipynb so tests pass

### LINKED ISSUE #342: Diffusion map notebook tests fail
See here for the con job failure: https://github.com/MDAnalysis/UserGuide/actions/runs/6858672704/job/18652083982

Looks like the eigenvector matrix is switching from negative to positive?

pgbarletta:
Fixes #342 

Error showed up on 14/11, after passing on 10/11. The only commit on the main MDA repo during that time was on the 11/11, which [refactored distance calculation](https://github.com/MDAnalysis/mdanalysis/commit/f4005db383f3f4c446433fb32156236f29f251ba).
 ``DiffusionMap`` uses ``DistanceMatrix``, which I assume uses the distances functions exposed through cython.

Given that the sign change in an eigenvector is not relevant and that we found (or at least I think so), where the change came from, I think we can just update the stored notebook so the test pass again.

----
:books: Documentation preview :books:: https://mdanalysisuserguide--345.org.readthedocs.build/en/345/


IAlibay:
The slight modification to the distance backend _shouldn't_ have been anything else than cosmetic - we did get this issue popping up irregularly before, I wonder if it's some upstream change.

In any case - @RMeli @lilyminium as leads on documentation could one of you please triage this?

lilyminium (approved):
I agree, the sign change isn't relevant and updating notebooks seems best here. LGTM -- thank you @pgbarletta!

RMeli:
Thanks @lilyminium for taking care of this. The sign change is not relevant, and can be caused by many things, such as changes in the underlying eigensolver implementation.

IAlibay:
Maybe to clarify my comment re: sign change & upstream - it's not that I don't agree that it's not relevant, but rather that it's probably not going to consistently be one way or the other (edit: because it's not behaviour we introduced). I.e. it may be worth considering if we'll be here with another failure next week just because things flipped.

pgbarletta:
Ok, if it breaks again, next time I'll ignore the output of that cell.

Il dom 10 dic 2023, 07:59 Irfan Alibay ***@***.***> ha
scritto:

PR #346, opened 2023-12-01
Fix by replacing tau_franes with tau_times

### LINKED ISSUE #344: An error in the example on the MDAnalysis user guide webpage
When calculating the Hydrogen bond autocorrelation on the MDAnalysis user guide webpage, there is an error in the code block for fitting the Hydrogen bond time constant. The page is [this](https://userguide.mdanalysis.org/stable/examples/analysis/hydrogen_bonds/hbonds-lifetimes.html). 

The ninth code block on the website is written as:
**params, fit_t, fit_ac = fit_biexponential(tau_frames, hbond_lifetime)**
But it should actually be:
**params, fit_t, fit_ac = fit_biexponential(tau_times, hbond_lifetime)**

Since the fitting uses this formula: A * np.exp(-t / tau1) + B * np.exp(-t / tau2), the units of t and tau1 should be the same, so it should be tau_times. However, because the dt of the simulation trajectory in the example is 1ps, it causes tau_frames and tau_times to be the same, making the time constant calculation results of the example appear normal.

The reason I did not use the maintenance method provided in your user guide is that this method is too complicated for chemical practitioners (as opposed to computer practitioners).

pgbarletta:
Fixes #344 

The user pointed out exactly why this notebook worked even though it was conceptually wrong.

----
:books: Documentation preview :books:: https://mdanalysisuserguide--346.org.readthedocs.build/en/346/


lilyminium (approved):
LGTM. Thanks for the fix @pgbarletta!

FYI, if some time has passed and you don't get a response, please do feel free to ping me or Rocco directly -- it can be easy for things to slip through the cracks during busy times! (I was just at back-to-back conferences and getting covid *yet again*)

Issue #354, opened 2024-01-02
Environment cache is not clearing properly


IAlibay:
Follow up from: https://github.com/MDAnalysis/UserGuide/pull/273
See: https://github.com/MDAnalysis/UserGuide/pull/353#issuecomment-1874281718

Looks like the cache isn't clearing at all, so we're hitting 3+ months on CI using some old cached environment.

PR #355, opened 2024-01-02
Turn off caching

IAlibay:
Temporary patch for #354 

There's a cache clearing solution here, but for now this will at least avoid us using super old environments.

----
📚 Documentation preview 📚: https://mdanalysisuserguide--355.org.readthedocs.build/en/355/


lilyminium (approved):
Whoops, thank you. I'm just going to approve and merge this since there's some circular PR dependencies (need to pick up this for #356, etc...) and it seems to make sense!

Issue #357, opened 2024-01-02
nbval errors on stderr output changes


IAlibay:
See: #356

FAILED polymer.ipynb::Cell 1
FAILED clustering_ensemble_similarity.ipynb::Cell 3
FAILED clustering_ensemble_similarity.ipynb::Cell 11
FAILED clustering_ensemble_similarity.ipynb::Cell 15
FAILED clustering_ensemble_similarity.ipynb::Cell 18
FAILED convergence.ipynb::Cell 2
FAILED convergence.ipynb::Cell 8
FAILED dimension_reduction_ensemble_similarity.ipynb::Cell 3
FAILED dimension_reduction_ensemble_similarity.ipynb::Cell 10
FAILED dimension_reduction_ensemble_similarity.ipynb::Cell 20
FAILED harmonic_ensemble_similarity.ipynb::Cell 3

IAlibay:
Do we actually need to do a stderr check here? Could we just skip it completely?

lilyminium:
Makes sense to me!

IAlibay:
Very quick attempt at doing this in https://github.com/MDAnalysis/UserGuide/actions/runs/7388882737/job/20100603113?pr=358 doesn't seem to work. Someone might be better suited to take over here.

PR #359, opened 2024-01-02
include date in cache key

mikemhenry:
This will cache the env + the package downloads. The date is included in the cache key so it will never be older than a day. This will still save some time/carbon when running CI, but will ensure that scheduled runs will get a "fresh" env.

----
📚 Documentation preview 📚: https://mdanalysisuserguide--359.org.readthedocs.build/en/359/

### DIFF
--- a/.github/workflows/gh-ci-tests.yaml
+++ b/.github/workflows/gh-ci-tests.yaml
@@ -29,12 +29,19 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Get current date
+        id: date
+        run: echo "date=$(date +%Y-%m-%d)" >> "${GITHUB_OUTPUT}"
+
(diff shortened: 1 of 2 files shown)


IAlibay (approved):
Thanks @mikemhenry !

I've taken the liberty to add in the same fix to gh-ci-tests, hopefully that's doing it right.

mikemhenry:
Ah when I looked at #355 it was just that file -- but your changes look correct!

IAlibay:
> Ah when I looked at #355 it was just that file -- but your changes look correct!

Yeah looks like I goofed in #355 🙈 sorry!

IAlibay:
CI failure is expected, see #357

IAlibay:
Thanks for the quick solution @mikemhenry !

PR #363, opened 2024-02-11
Add LGPL and CC-BY-SA LICENSE

### LINKED ISSUE #134: License for the user guide
So I might not be looking at the right place, but as far as I can tell the user guide doesn't directly declare any type of license.

I guess the assumption is that it inherits MDA's GPLv2 license?

Given that the user guide is a really good teaching resource, I think it might be good to make it clear as to what the license is so that folks can be more confident in using parts of it in derivative works where possible.

RMeli:
Fix #134.

@MDAnalysis/coredevs, thoughts? 

----
📚 Documentation preview 📚: https://mdanalysisuserguide--363.org.readthedocs.build/en/363/

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,504 @@
+                  GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 2.1, February 1999
+
+ Copyright (C) 1991, 1999 Free Software Foundation, Inc.
+ 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+[This is the first released version of the Lesser GPL.  It also counts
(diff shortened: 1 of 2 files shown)


IAlibay:
The code bits can't be LGPL if they are derivatives of GPL, you'll have to wait until relicense

RMeli:
> you'll have to wait until relicense

Yep, this was my assumption. It would be annoying to have to re-license here too.

orbeckst (changes requested):
Looks good but I have to block until the MDA relicense to LGPL is done.

RMeli:
Pinging since now the relicensing is done.

orbeckst (approved):

orbeckst:
@lilyminium , as the original author of the UG, could you please als ok this license assignment?

lilyminium (approved):
LGTM!! Sorry for the holdup.

PR #387, opened 2024-10-11
Add reference to gromacs manual for xtc

RubenChM:
----
📚 Documentation preview 📚: https://mdanalysisuserguide--387.org.readthedocs.build/en/387/

PR #388, opened 2024-10-11
Use setup-python instead of setup-micromamba for pre-commit CI

RMeli:
----
📚 Documentation preview 📚: https://mdanalysisuserguide--388.org.readthedocs.build/en/388/

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -24,19 +24,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-
-    - name: setup micromamba
-      uses: mamba-org/setup-micromamba@v1
+    - uses: actions/setup-python@v3
       with:
-        environment-file: environment.yml
-        environment-name: mda-user-guide
-        create-args: >-
-          pip
-        condarc: |
-          channels:
-            - conda-forge
-            - plotly
-
-    - uses: pre-commit/action@v3.0.0
+        python-version: '3.x'
+    - uses: pre-commit/action@v3.0.1
       with:
         extra_args: --all-files


RMeli:
We had a failure in #387. Trying to see if using `setup-python` works.

Others encountered the same issue https://github.com/pre-commit/action/issues/214, https://github.com/pre-commit/action/issues/213, https://github.com/pre-commit/action/issues/210. Honestly, the replies from the maintainer on these issues makes me want to steer clear of the `pre-commit` GitHub action.

Additionally:
> this action is in maintenance-only mode and will not be accepting new features.
> generally you want to use [pre-commit.ci](https://pre-commit.ci/) which is faster and has more features.

Maybe we should consider moving to `pre-commit.ci`? @jandom?

.github/workflows/pre-commit.yaml
```diff
@@ -24,19 +24,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-
-    - name: setup micromamba
-      uses: mamba-org/setup-micromamba@v1
+    - uses: actions/setup-python@v3
       with:
-        environment-file: environment.yml
```
orbeckst, line 31: I assume that the full UG env is not needed for the pre-commit action and that's why you changed to a simple python env?

orbeckst (approved):
I know very little about the background here. I am providing a "LGTM" because it looks sensible to use a simple python setup to run pre-commit and because you are of the opinion that it will improve the CI and because it didn't visibly break anything.

If someone with more experience could have a look then that would be good, of course, but I ultimately trust that you know what you're doing.

RMeli:
I could not find much information on _why_ this is happening now. But it seems that the action is only in maintenance mode, and the maintainer is very quick at closing issues. In the ones outlining this problem (linked above, _links were wrong and I just updated them_), replies are of the sort

> you are the third or fourth person to not search before creating an issue for this specific thing. please do better!

or

> skill issue imo. it is in the readme you just didn't follow it or read the output of your own job

So not much information there. The README uses the `setup-python` action to setup Python for pre-commit, so that's what I went for and it seems to work as expected.

.github/workflows/pre-commit.yaml
RMeli: Yes. The pre-commits builds an environment with its own tools (with the versions specified in the pre-commit configuration).

orbeckst: All makes sense to me, thanks for explaining.

orbeckst:
The maintainer comments on the pre-commit issues are not particularly welcoming. I don't quite know what would be better for a "maintenance-only" project: no response at all or harsh responses + shutting down issues. 🤷

IAlibay (approved):
Sounds like the precomit action is doing something reasonably nefarious.. I do agree that moving away from it would be good.

PR #411, opened 2025-03-10
Update 2.8.0 release notes to include guessing

lilyminium:
----
📚 Documentation preview 📚: https://mdanalysisuserguide--411.org.readthedocs.build/en/411/

### DIFF
--- a/doc/source/releases.md
+++ b/doc/source/releases.md
@@ -17,7 +17,7 @@ Supported Python versions:
 See the [CHANGELOG](https://github.com/MDAnalysis/mdanalysis/blob/release-2.8.0/package/CHANGELOG) and our [release blog post](https://www.mdanalysis.org/blog/#mdanalysis-2.8-is-out) for more details.
 
 * The MDAnalysis package license has changed from GPLv3+ to LGPLv3+ and all code contributions are now under LGPLv2.1+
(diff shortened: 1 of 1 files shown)


IAlibay (commented):
@lilyminium aren't these auto generated? Won't they just get overwritten the next time we update?

IAlibay:
Here's the auto-generation script: https://github.com/MDAnalysis/UserGuide/blob/develop/doc/source/scripts/gen_release_notes.py

lilyminium:
Yeah I have kind of mixed feelings about this script since IMO the user guide changelog should reflect changes in the user guide too, and moving the guessers to a top-level might mess up someone's links; was thinking of updating the script to try to just add new sections on top instead of re-writing the file. I don't feel strongly either way though so happy to close this PR if you reckon it's too much effort.

IAlibay:
> the user guide changelog should reflect changes in the user guide too,

Not sure I understand sorry. That page was meant to be a user-facing version of the core library changelog. I remember adding that as a request from other coredevs that wanted something that was actually readable.

I don't _think_ it's intended to be the changelog for the userguide?

lilyminium:
Oh yes I remember, thanks for the reminder! Apologies, I think I got all mixed up here -- trying to do some of this during meetings. I'll close this.

Issue #96, opened 2020-07-25
selections: icodes + empty select


lilyminium:
Document icodes + empty fnmatch (`[!?]`) in selection strings

Issue #42, opened 2020-02-13
Salt bridge analysis


lilyminium:
[An explicit example of salt bridge analysis.](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/mdnalysis-discussion/isqHNXt_TTg)

orbeckst:
@ianmkenney what code do you use for saltbridge analysis?

Issue #15, opened 2019-09-30
syntax highlighting missing in UserGuide


orbeckst:
The code snippets are not syntax highlighted, eg Cell 1 in https://www.mdanalysis.org/UserGuide/quickstart#Quick-start-guide

PR #430, opened 2025-10-10
Add UserGuide documentation for `IMDReader` 

### LINKED ISSUE #427: Add UserGuide documentation for `IMDReader` - reading streaming data with IMDv3 protocol
**User story**

As a user familiar with MDAnalysis, but unaware of the new IMDReader-IMDv3 feature, I would want to easily discover and learn how to use the streaming reader with the new IMDv3 protocol support without having to dig through API docs.

**Additional context**

MDAnalysis is [adding (recently added)](https://github.com/MDAnalysis/mdanalysis/pull/4923) support for real-time streaming via the IMDv3 protocol through the new IMDReader, but this feature is only documented in the API docs. Most users read the UserGuide to learn about MDA capabilities, so they're unlikely to discover this functionality. The current API docs assumes that users already know about streaming and focuses on technical details rather than helping them understand what it is, and when and how to use it.

This creates a discoverability gap for a feature that could change how users approach simulation analysis - enabling real-time monitoring, and in-situ analysis.

### 1. **Introduction: What is streaming?**
- Brief explanation of IMDv3 protocol vs traditional file-based analysis
- Benefits: real-time monitoring, in-situ analysis et al.

### 2. **When to use streaming?**
- Real-time quality control and monitoring (Long-running simulations where early feedback is valuable)
- Adaptive sampling, selection
- Interactive visualization during simulation
- Avoid large data storage - in-situ analysis

### 3. **Installation and setup**
- Installing imdclient: `pip install imdclient`
- MD engine support status and configuration:
  - GROMACS: example .mdp settings and mdrun flags
  - LAMMPS: configuration examples
  - NAMD: configuration examples
- Links to engine-specific documentation

### 4. **Basic streaming example - how to use `IMDReader`**
Basic example using streaming, something like ---
```python
import MDAnalysis as mda

# Connect to running simulation
u = mda.Universe("topol.tpr", "imd://localhost:8889")

# Streaming loop
for ts in u.trajectory:
    print streaming_info
    
    # Your analysis code here
    if some_stopping_condition:
        break
```

### 5. **Additional features**
Features that are available through IMDClient via the IMDReader
- `continue_after_disconnect`
- `buffer_size`
- pause/resume? (not available directly)

### 6. **Analysis workflows**
- Integration with MDAnalysis analysis modules
(which analysis methods work/don't with IMDv3 MDA (limitations from StreamReaderBase)
- Examples?

### 7. **Complete working example**
End-to-end workflow: GROMACS/NAMD/LAMMPS setup and streaming --> IMDReader --> visualization/analysis

### 8. **Current limitations**
- Limitations with accessing data
  - implications for analysis usage
  - (from streamreaderbase)

Above information needs to be updated as `imd.rst` under `doc/source/formats/reference`

amruthesht:
The following PR address Issue #427 and adds UserGuide documentation for the `IMDReader` via `imd.rst` and a tutorial/example `simulation_imd.ipynb`.

The template for the documentation followed is deatiled in the original Issue #427

----
📚 Documentation preview 📚: https://mdanalysisuserguide--430.org.readthedocs.build/en/430/


orbeckst:
@ljwoods2 could you also have a quick look please?

doc/source/examples/other/streaming_imd.ipynb
```diff
@@ -0,0 +1,325 @@
+{
```
ljwoods2, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    # !pip install imdclient&gt;=0.2.2
</pre><p>Don't we want people to install with Conda for consistency?</p><p> </p><p>Also, will people be running this notebook, or reading it as a guide? I think this cell can potentially be removed, seems like a bit of overkill</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MDAnalysis/UserGuide/pull/430/discussion/#comment-2421498717'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

ljwoods2, line 1: <p>Why is IMD-coords essential for GROMACS? Will the stream fail out without it? Or do you just mean its essential for this example?</p><p> </p><p>Maybe it's worth specifying exactly what's necessary for this example: you need to have coordinates turned on regardless of simulation engine, you need to have freq set to 1, and you need to turn off coordinate unwrapping (so that there are consistent configs across engines)</p><p> </p><p>Also, should this link the docs for each of the simulation engines?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MDAnalysis/UserGuide/pull/430/discussion/#comment-2421498724'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

ljwoods2, line 1: <pre class="ql-syntax" spellcheck="false">Line #42.    # simple_imd_analysis()
</pre><p> </p><p>Why is this commented out? Will github run it automatically as a part of docstring tests if it's uncommented?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MDAnalysis/UserGuide/pull/430/discussion/#comment-2421498740'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

ljwoods2, line 1: <pre class="ql-syntax" spellcheck="false">Line #41.                potential_energy = -1000 + np.random.normal(0, 50)
</pre><p>Not all simulation engines implement energy packet, so maybe remove this?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MDAnalysis/UserGuide/pull/430/discussion/#comment-2421498746'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

ljwoods2, line 1: <pre class="ql-syntax" spellcheck="false">Line #51.                    ax1.clear()
</pre><p>Just double checking, did you test this method of live updating plot locally? Looks a bit different than what we did in workshop</p><p> </p><p><a href="https://github.com/MDAnalysis/imd-workshop-2024/blob/main/activity/graph_utils.py" target="_blank">https://github.com/MDAnalysis/imd-workshop-2024/blob/main/activity/graph_utils.py</a></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MDAnalysis/UserGuide/pull/430/discussion/#comment-2421498754'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+==================================================================
+IMD (Data streamed via Interactive Molecular Dynamics protocol v3)
+==================================================================
+
+.. include:: classes/IMD.txt
+
+Real-time streaming of simulation data between molecular dynamics engines and receiving clients can be achieved using IMD protocols like IMDv2 and IMDv3. The :class:`~MDAnalysis.coordinates.IMD.IMDReader` implements the IMDv3 protocol, enabling live streaming of ongoing simulation data.
+
+.. note::
+   MDAnalysis supports **IMDv3 only**, which provides continuous, gap-free streaming and is implemented in modern versions of GROMACS, LAMMPS, and NAMD. IMDv2, while widely available, was designed primarily for visualization and allows gaps in the data stream.
```
ljwoods2:
```suggestion
   MDAnalysis supports **IMDv3 only**, which provides continuous, gap-free streaming and is implemented in modern versions of GROMACS, LAMMPS, and NAMD. IMDv2, while widely available, was designed primarily for visualization and doesn't enforce a consistent number of integration steps between transmitted frames
```

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+What is Streaming?
+==================
+
+Streaming involves processing data in real-time as it is generated, rather than storing it for later analysis. In molecular dynamics, this means sending simulation data to a client on-the-fly while the simulation is running, without writing large trajectory files to disk.
+
+This can be achieved through a TCP/IP socket connection between the simulation engine and receiving client, transmitting coordinates, velocities, forces, energies, and timing information using the IMDv3 protocol.
```
ljwoods2:
```suggestion
In IMDv3, this is achieved through a TCP/IP socket connection between the simulation engine and receiving client, transmitting coordinates, velocities, forces, energies, and timing information.
```

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
+.. -*- coding: utf-8 -*-
+.. _IMD-format:
+
+==================================================================
+IMD (Data streamed via Interactive Molecular Dynamics protocol v3)
+==================================================================
+
+.. include:: classes/IMD.txt
+
+Real-time streaming of simulation data between molecular dynamics engines and receiving clients can be achieved using IMD protocols like IMDv2 and IMDv3. The :class:`~MDAnalysis.coordinates.IMD.IMDReader` implements the IMDv3 protocol, enabling live streaming of ongoing simulation data.
```
ljwoods2:
```suggestion
IMDv2 and IMDv3 enable real-time streaming of simulation data between molecular dynamics engines and receiving clients. The :class:`~MDAnalysis.coordinates.IMD.IMDReader` implements the IMDv3 protocol.
```

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+======================
+
+Required Dependencies
+---------------------
+
+The IMDReader requires the ``imdclient`` package:
+
+.. code-block:: bash
+
+   pip install imdclient
```
ljwoods2: I think we encourage conda/mamba installation

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+       # Check for simulation artifacts
+       if previous_positions is not None:
+           displacement = np.linalg.norm(current_positions - previous_positions, axis=1)
+           max_displacement = np.max(displacement)
+           
+           if max_displacement > 10.0:  # Atoms moved > 10 Å in one step
+               print(f"WARNING: Large displacement detected at {ts.time} ps: {max_displacement:.2f} Å")
+       
+       # Monitor energies if available
+       if 'potential' in ts.data:
```
ljwoods2: I think the key is "potential_energy" 

https://github.com/Becksteinlab/imdclient/blob/07069e1ff08d1488936c61e52f922116ecddf210/imdclient/IMDClient.py#L956

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+Buffer Management
+-----------------
+
+For compute-intensive analysis, increase the buffer size to reduce communication overhead:
+
+.. code-block:: python
+
+   # Larger buffer for better performance
+   u = mda.Universe("topol.tpr", "imd://localhost:8889", buffer_size=50*1024*1024)
```
ljwoods2, line 236: Specify unit, these are bytes

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+       u.trajectory.close()
+
+Available Data
+--------------
+
+The IMDReader provides access to additional simulation data through ``ts.data``:
+
+* ``dt``: Time step size in picoseconds
+* ``step``: Current simulation step number  
+* Energy terms: ``potential``, ``total``, etc. (engine-dependent)
```
ljwoods2: I think these keys are wrong, see:

https://github.com/Becksteinlab/imdclient/blob/07069e1ff08d1488936c61e52f922116ecddf210/imdclient/IMDClient.py#L956

I think we actually merged them into MDA develop docstring for IMDReader incorrectly:

https://imdclient.readthedocs.io/en/latest/protocol_v3.html#energies

ljwoods2: Also, are these engine-dependent? It's engine depended whether the energy packet is implemented, but the exact energy keys should be consistent whenever the packet is implemented, right?

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+   for ts in u.trajectory:
+       print(f"Step {ts.data.get('step')}: dt={ts.data.get('dt')} ps")
+       
+       # Available energy terms vary by MD engine
+       for key, value in ts.data.items():
+           if key not in ['dt', 'step']:
+               print(f"  {key}: {value}")
+
+Multiple Client Connections
```
ljwoods2: There's a ton of edge cases when using multiple clients that we don't test, like when one client tries to change blocking behavior, so I think we should just say it isn't recommended

Discussion here

https://github.com/Becksteinlab/imdclient/pull/114 

Maybe should update MDA main codebase docs to say this as well:

https://github.com/amruthesht/mdanalysis/blob/ea85bb510f684ef1f6048ad0a50219665298cbb3/package/MDAnalysis/coordinates/IMD.py#L108

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+.. important::
+   Even when multiple connections are supported, each receives an independent data stream. Different clients may receive different data depending on the engine configuration.
+
+Integration with MDAnalysis Tools
+=================================
+
+Most MDAnalysis analysis classes work with streaming data, but some limitations apply:
+
+**Compatible Analysis**
```
ljwoods2: I think this should start by outlining what kinds of analyses will and won't work before jumping into distance and contact analysis

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+**Practical Considerations**
+
+.. code-block:: python
+
+   # This WILL work - forward iteration
+   for ts in u.trajectory:
+       analysis_data.append(calculate_something(ts))
+   
+   # This will NOT work - random access
+   ts = u.trajectory[10]  # TypeError
```
ljwoods2: From reading docstring of `__getitem__` of StreamReaderBase, I think these should both be ValueErrors

``` 
def __getitem__(self, frame):
        """Return an iterator for slicing a streaming trajectory.

        Parameters
        ----------
        frame : slice
            Slice object. Only the step parameter is meaningful for streams.

        Returns
        -------
        FrameIteratorAll or StreamFrameIteratorSliced
            Iterator for the requested slice.

        Raises
        ------
        TypeError
            If frame is not a slice object.
        ValueError
            If slice contains start or stop values.
```

If it does throw a TypeError, something is wrong...

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+   
+   # This will NOT work - backwards iteration  
+   for ts in u.trajectory[::-1]:  # ValueError
+       pass
+   
+   # This will NOT work - restarting iteration
+   for ts in u.trajectory:
+       break
+   for ts in u.trajectory:  # Won't start from beginning
```
ljwoods2: This suggest it will start from somewhere else other than the beginning, won't it just fail?

ljwoods2 (changes requested):

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+   
+   # Connect to the simulation
+   u = mda.Universe("topol.tpr", "imd://localhost:8889", buffer_size=10*1024*1024)
+   
+   # Streaming analysis loop
+   for ts in u.trajectory:
+       print(f"Time: {ts.time:.2f} ps, Step: {ts.data.get('step', 'N/A')}")
+       
+       # Your analysis code here
+       selected_atoms = u.select_atoms("protein and name CA")
```
tylerjereddy: to clarify -- is there something about IMD that mandates repeating the selection of the `AtomGroup` in the loop, or can this be hoisted up for efficiency as usual?

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+       center_of_mass = selected_atoms.center_of_mass()
+       print(f"Protein COM: {center_of_mass}")
+       
+       # Optional: break on some condition
+       if ts.time > 1000:  # Stop after 1000 ps
```
tylerjereddy, line 183: to clarify: this *only* stops the analysis, not the running simulation/engine the analysis is connected to?

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+   u = mda.Universe("system.tpr", "imd://localhost:8889")
+   
+   for ts in u.trajectory:
+       # Distance calculations work normally
+       protein = u.select_atoms("protein")
+       rg = protein.radius_of_gyration()
+       
+       # Contact analysis
+       selection1 = u.select_atoms("resid 1-10")
+       selection2 = u.select_atoms("resid 50-60")
```
tylerjereddy: hoist selections if IMD allows it?

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+* **Forward-only**: Can only iterate through frames as they arrive
+* **Single-use**: Cannot restart iteration once the stream is consumed
+* **No trajectory length**: Total frame count unknown until simulation ends
+* **No independent copies**: Cannot create multiple reader instances for the same stream
+
+**Analysis Constraints**
+
+* **No timeseries methods**: Cannot use ``trajectory.timeseries()``
+* **No bulk operations**: Cannot extract all data at once
+* **Limited multiprocessing**: Cannot split across processes
```
tylerjereddy, line 371: Is there a concrete example of something you can do with `multiprocessing` with "vanilla" MDAnalysis that doesn't work with IMD? I would have figured that if you can get data into a NumPy array you're probably "ok" to split off additional processes if you want to.

doc/source/formats/reference/imd.rst
ljwoods2: Nothing unique about IMD here, can be changed

doc/source/formats/reference/imd.rst
ljwoods2: Since this break is the end of the script, it will trigger the IMDClient to cleanup via atext

https://github.com/Becksteinlab/imdclient/blob/07069e1ff08d1488936c61e52f922116ecddf210/imdclient/IMDClient.py#L146

That cleanup might put the simulation engine in a waiting state for another client connection, but won't kill it.

Whether it makes the simulation engine wait or not depends on 1. how the simulation engine was configured and 2. whether the IMDClient overrode the simulation engine's waiting behavior

`continue_after_disconnect` kwarg is the IMDClient switch for changing waiting behavior

https://imdclient.readthedocs.io/en/latest/api.html

doc/source/formats/reference/imd.rst
ljwoods2: By vanilla MDA, does that exclude analysis classes with dask/multiprocessing backend?

You could copy the entire traj into np array as it goes, then create universe with traj from array, but that would defeat the purpose I think 

And you can use multiprocessing/parallelism on a single frame / sliding window of frames, I know @HeydenLabASU has tried some windowing and then using parallel algorithms to process an i.e. 200 frame window

I guess maybe this should be changed to say that it only works with the serial analysis backend

doc/source/formats/reference/imd.rst
tylerjereddy: I wonder if the need to look out for this/think about this might be sensible to mention somewhere inline here then, if only to refer to more detailed docs for someone who may be copy/pasting the sample code as a template.

doc/source/formats/reference/imd.rst
tylerjereddy: >  that would defeat the purpose I think

I guess you could "buffer"/store a few frames in an array that way without doing IO, and possibly only for a select subset of particles, which does seem within the remit/spirit of what IMD enables, or possibly even one of the more interesting use cases. I think that's what you're saying Matthias is doing.

I might guess that accessing multiple frames in parallel is a problem for the same reason that random access is not available, and that some analyses that require multiple passes and divide work over random-access frame access are also banned, so yeah just trying to think of cases that block multiprocessing usage that aren't already covered by the elementary limitations mentioned. Maybe the multiple clients in different processes thing, but I think you also already ruled them out in comments above, whether in different processes or not.

doc/source/examples/other/streaming_imd.ipynb
orbeckst: ~~I don't think we need package installation here. imdclient will always be installed when MDA is installed.~~ imdclient will need to be installed separately. However, I'd mention this fact and then link to installation instructions instead of repeating instructions that may change over time.

In general, always link to authoritative instructions; you're _the_ expert on the matter so you should know what the authoritative sources are.

orbeckst: <p>Link to </p><ul><li>MD engine official web pages</li><li>imdclient docs <a href="https://imdclient.readthedocs.io/en/stable/usage.html" target="_blank">https://imdclient.readthedocs.io/en/stable/usage.html</a></li></ul>

orbeckst: <p>The notebook needs to be primarily a readable document, don't expect people to edit it. Thus, it should show _working code_, so avoid "un-commenting something so that it works".</p><p> </p><p>If there are issues with it being run, comment on the PR and then let's look for a solution.</p>

orbeckst: <p>That's true but for the ones that do, we should show how one can get the information.</p><p> </p>

orbeckst, line 1: <p>Include links to the authoritative docs — imdclient and IMDReader in the main docs.</p><p> </p><p>The imdclient package should be installed as a dependency of MDAnalysis (in 2.10.0...IIRC(</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MDAnalysis/UserGuide/pull/430/discussion/#comment-2433986956'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

orbeckst, line 1: <p>For all codes:</p><p> </p><p>1) Do we want to send IMD data every integrator step in the example? It's ok to do so but write some text above that says explicitly that we are doing it in the example but in production simulations this may be every 5 or every 10 steps.</p><p> </p><p>For GROMACS:</p><p> </p><p>2) Let's model best practice and encourage everyone to send time and box information so that this is <em>always</em> available in MDAnalysis:</p>

<pre class="ql-syntax" spellcheck="false">IMD-time         = Yes         ; Send time information
...
IMD-box          = Yes         ; Send box dimensions (recommended)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MDAnalysis/UserGuide/pull/430/discussion/#comment-2433986964'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

orbeckst, line 1: <p>Explain <em>what</em> analysis you're performing. Link to <a href="https://docs.mdanalysis.org/stable/documentation_pages/core/groups.html#MDAnalysis.core.groups.AtomGroup.radius_of_gyration" target="_blank">https://docs.mdanalysis.org/stable/documentation_pages/core/groups.html#MDAnalysis.core.groups.AtomGroup.radius_of_gyration</a> (or somewhere in the User Guide) for background.</p><p> </p><p>Also explain that **as an example** you only process the first 10 frames but normally you'd read all frames.</p><p> </p><p>You should also discuss</p><ul><li>skipping (<code>trajectory[::5]</code>)</li><li>limitations (no start/stop)</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MDAnalysis/UserGuide/pull/430/discussion/#comment-2433986966'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

orbeckst, line 1: <p>I don't like the structure of the code in that the contents of the <code>try</code> block are the whole code. If you want to protect code then make a function out of the inner block. However, I would just not use the try/except and just state that the example requires a running simulation.</p><p> </p><p>(Additionally, I don't like catching all exceptions and then just printing an error — this is bad practice because errors can slip through.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MDAnalysis/UserGuide/pull/430/discussion/#comment-2433986973'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

orbeckst, line 1: <pre class="ql-syntax" spellcheck="false">Line #21.            frame_count = 0
</pre><p>In the text, explain that one of the limitations is that one cannot use <code>trajectory[:10]</code> to make clear why you're keeping track of frames. (You can link to <a href="https://docs.mdanalysis.org/dev/documentation_pages/coordinates/base.html#MDAnalysis.coordinates.base.StreamReaderBase " target="_blank">StreamReaderBase</a> for more details on limitations).</p><p> </p><p>By the way, couldn't you use <code>ts.frame</code>? That would at least not be quite as clunky.   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MDAnalysis/UserGuide/pull/430/discussion/#comment-2433986983'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

orbeckst, line 1: <pre class="ql-syntax" spellcheck="false">Line #15.        try:
</pre><p>Same comment as above: kind of ugly to have so much code inside try/except. </p><p> </p><p>I'd rewrite the code block inside the try block as a function and then keep the existing try/except structure to demonstrate the overall control structure. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MDAnalysis/UserGuide/pull/430/discussion/#comment-2433986993'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

orbeckst, line 1: <pre class="ql-syntax" spellcheck="false">Line #41.                potential_energy = -1000 + np.random.normal(0, 50)
</pre><p>Why do we fake it? Instead we should get it from the energies packet. You just need to say above that one needs to enable sending of energies and then pull the correct data out of the <code>ts.data</code>. This would be a good example for showing how to do it, don't squander the opportunity.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MDAnalysis/UserGuide/pull/430/discussion/#comment-2433986997'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

orbeckst, line 1: <pre class="ql-syntax" spellcheck="false">Line #96.    # advanced_live_streaming()
</pre><p>No commented out code, see above.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MDAnalysis/UserGuide/pull/430/discussion/#comment-2433987003'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

doc/source/formats/reference/imd.rst
orbeckst, line 1: General comment: link to the primary sources (imdclient docs, IMDReader docs)

doc/source/formats/reference/imd.rst
orbeckst: You need to explain what the general purpose is (re-summarize paper, IMDReader, imdclient docs). Put yourself in the position of someone whose **first contact** with the idea is through the User Guide.  Answer the questions

* What is it?
* Why is it useful?
* Where can I learn more?

Also explain abbreviation "IMD". 

Recyle what you've already have written, e.g. from the "What is streaming" section, which is already great in answering the first question. You just need to taylor the initial 2-3 sentences so that they make sense on their own.

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+Real-time streaming of simulation data between molecular dynamics engines and receiving clients can be achieved using IMD protocols like IMDv2 and IMDv3. The :class:`~MDAnalysis.coordinates.IMD.IMDReader` implements the IMDv3 protocol, enabling live streaming of ongoing simulation data.
+
+.. note::
+   MDAnalysis supports **IMDv3 only**, which provides continuous, gap-free streaming and is implemented in modern versions of GROMACS, LAMMPS, and NAMD. IMDv2, while widely available, was designed primarily for visualization and allows gaps in the data stream.
+
+What is Streaming?
+==================
+
+Streaming involves processing data in real-time as it is generated, rather than storing it for later analysis. In molecular dynamics, this means sending simulation data to a client on-the-fly while the simulation is running, without writing large trajectory files to disk.
```
orbeckst: This is good!

doc/source/formats/reference/imd.rst
orbeckst: @ljwoods2 suggestion is an improvement (so I am adding it) but it's still too much jargon for an opening paragraph — see my comments above.

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+This can be achieved through a TCP/IP socket connection between the simulation engine and receiving client, transmitting coordinates, velocities, forces, energies, and timing information using the IMDv3 protocol.
+
+MDAnalysis's :class:`~MDAnalysis.coordinates.IMD.IMDReader` uses the `imdclient <https://imdclient.readthedocs.io/>`_ package and provides a familiar interface for reading streaming data, similar to other trajectory readers in MDAnalysis.
+
+When to Use Streaming?
```
orbeckst, line 39: very nice — answers my questions from the first paragraph: so summarize in the first paragraph at a high level what you're writing here.

doc/source/examples/other/streaming_imd.ipynb
orbeckst: This section appears to be a duplicate of what's in the imd.rst. Can you just omit all of it and just link?

It's basically impossible to keep documentation in sync so the best you can do is to only have it in a single place and link — **DRY**, don't repeat yourself.

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+**Adaptive sampling workflows**
+    Real-time analysis can guide simulation parameters or trigger enhanced sampling methods.
+
+**Interactive research**
+    Immediate feedback allows researchers to make informed decisions about continuing, modifying, or terminating simulations.
+
+**Storage-constrained environments**
+    Analyze data as it's generated without storing large trajectory files.
+
+Installation and Setup
```
orbeckst, line 56: Explain that imdclient is an _optional_ package and has to be installed separately from MDAnalysis, either with mamba or pip.

I'd just condense the whole section and link to the authoritative https://imdclient.readthedocs.io/en/stable/getting_started.html#installation

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+======================
+
+Required Dependencies
+---------------------
+
+The IMDReader requires the ``imdclient`` package:
```
orbeckst: Link imdclient to its docs.

doc/source/formats/reference/imd.rst
orbeckst: yes!

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+.. code-block:: bash
+
+   pip install imdclient
+
+.. note::
+   MDAnalysis requires ``imdclient >= 0.2.2`` for its current implementation.
+
+MD Engine Configuration
```
orbeckst, line 64: Definitely link to the imdclient docs that say more or less the same thing.

orbeckst: I'd also suggest to _not_ repeat this configuration in the notebook and just link from the notebook to here.

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+-----------------------
+
+We provide below example configurations for enabling IMDv3 streaming in popular MD engines.
+
+**GROMACS**
+
+Add IMD settings to your ``.mdp`` file:
```
orbeckst: Add text that _explains_ what these settings do, or at minimum highlight the important settings.

Definitely provide a link to the imdclient docs (which should be the authoritative source).

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+.. code-block:: text
+
+    ; IMD settings
+    IMD-group        = System
+    IMD-version      = 3
+    IMD-nst          = 1
+    IMD-time         = No
```
orbeckst:
```suggestion
    IMD-time         = Yes
```

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+    IMD-coords       = Yes
+    IMD-vels         = No
+    IMD-forces       = No
+    IMD-box          = No
```
orbeckst: Being able to do this was almost the whole point of IMDv3
```suggestion
    IMD-box          = Yes
```
(+ velocities)

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+.. note::
+   MDAnalysis requires ``imdclient >= 0.2.2`` for its current implementation.
+
+MD Engine Configuration
+-----------------------
+
+We provide below example configurations for enabling IMDv3 streaming in popular MD engines.
+
+**GROMACS**
```
orbeckst: Make this a heading of the right level.

It's bad practice to use emphasis markup instead of semantic markup for structuring documents. Not only does it make formatting inconsistent (using styles) but it also makes documents hard to read on screen readers and thus decreases the accessibility dramatically.

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+
+
+Run with IMD enabled:
+
+.. code-block:: bash
+
+    gmx mdrun -v -nt 4 -imdwait -imdport 8889
+
+**LAMMPS**
```
orbeckst: heading

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+Use the IMD fix in your input script:
+
+.. code-block:: text
+
+    # IMD setup
+    fix ID group-ID imd <port> trate <frequency> version 3 unwrap <on/off> time <on/off> box <on/off> coordinates <on/off> velocities <on/off> forces <on/off>
+
+Run your LAMMPS simulation as usual.
+
+**NAMD**
```
orbeckst: heading

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+   pip install imdclient
+
+.. note::
+   MDAnalysis requires ``imdclient >= 0.2.2`` for its current implementation.
+
+MD Engine Configuration
+-----------------------
+
+We provide below example configurations for enabling IMDv3 streaming in popular MD engines.
```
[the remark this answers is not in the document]
orbeckst: To make this section different from the authoritative https://imdclient.readthedocs.io/en/stable/usage.html#configuring-the-simulation-engine-for-imdv3 I suggest that you give an _actual example_ configuration for each engine with real example values and not place holders like `<yes/no>`.

Imagine that students (and LLMS!) will use this information in order to get started.

If you were a PI, what values would you want your student to use? Put these in and add a sentence or two explaining your reasoning.

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+.. code-block:: text
+
+    ; IMD settings
+    IMD-group        = System
+    IMD-version      = 3
+    IMD-nst          = 1
+    IMD-time         = No
+    IMD-coords       = Yes
+    IMD-vels         = No
```
orbeckst: In the example also stream velocities.

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+    IMD-forces       = No
+    IMD-box          = No
+    IMD-unwrap       = No
+    IMD-energies     = No
```
orbeckst: and transmit energies

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+**GROMACS**
+
+Add IMD settings to your ``.mdp`` file:
+
+.. code-block:: text
+
+    ; IMD settings
+    IMD-group        = System
+    IMD-version      = 3
+    IMD-nst          = 1
```
orbeckst: perhaps not every step? what would you choose?

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+    # IMD setup
+    fix ID group-ID imd <port> trate <frequency> version 3 unwrap <on/off> time <on/off> box <on/off> coordinates <on/off> velocities <on/off> forces <on/off>
+
+Run your LAMMPS simulation as usual.
+
+**NAMD**
+
+Add IMD configuration to your NAMD configuration file:
+
+.. code-block:: text
```
orbeckst, line 134: Make the example equivalent to the GROMACS one.

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+.. code-block:: bash
+
+    gmx mdrun -v -nt 4 -imdwait -imdport 8889
+
+**LAMMPS**
+
+Use the IMD fix in your input script:
+
+.. code-block:: text
```
orbeckst, line 110: Make the example equivalent to the GROMACS one.

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+.. seealso::
+   For detailed engine-specific setup instructions, see the `imdclient simulation engine documentation <https://imdclient.readthedocs.io/en/latest/usage.html>`_.
+
+Basic Usage
+===========
+
+Connecting to a Running Simulation
+-----------------------------------
+
+Once your simulation is running with IMD enabled:
```
orbeckst, line 162: explain choice of buffer_size.

doc/source/formats/reference/imd.rst
orbeckst: @amruthesht move selection out of the loop to model best practices (it's bad for performance!)

doc/source/formats/reference/imd.rst
orbeckst: I'd remove the optional part from this example so that it can stand as a simple example on its own.

Then create another example where you discuss the issue with how breaking the loop may affect the running simulation. This is important but nuanced. If something is complicated, break it into smaller steps, that's basic pedagogy.

doc/source/formats/reference/imd.rst
orbeckst: For these examples I might not even build in safe guards but just state at the top that you're assuming that energy is transmitted.

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+   u = mda.Universe("system.tpr", "imd://localhost:8889")
+   
+   previous_positions = None
+   
+   for ts in u.trajectory:
+       current_positions = u.atoms.positions.copy()
+       
+       # Check for simulation artifacts
+       if previous_positions is not None:
+           displacement = np.linalg.norm(current_positions - previous_positions, axis=1)
```
orbeckst: The simple distance calculation fails on PBC crossing.

Use PBC-aware distance calculations with [`lib.distances.calc_bonds()`](https://docs.mdanalysis.org/stable/documentation_pages/lib/distances.html#MDAnalysis.lib.distances.calc_bonds). This will also demonstrate why it's important that we transfer box information!

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+       
+       previous_positions = current_positions
+
+Advanced Features
+=================
+
+Buffer Management
+-----------------
+
+For compute-intensive analysis, increase the buffer size to reduce communication overhead:
```
orbeckst: Link to imdclient docs that (hopefully) have more explanations about what the buffer does.

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+.. code-block:: python
+
+   # Larger buffer for better performance
+   u = mda.Universe("topol.tpr", "imd://localhost:8889", buffer_size=50*1024*1024)
+
+Connection Management
```
orbeckst, line 240: This is a good section.

If you have other try/except or specialized handling then you can collect it here.

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+   import MDAnalysis as mda
+   
+   try:
+       u = mda.Universe("topol.tpr", "imd://localhost:8889")
+       
+       for ts in u.trajectory:
+           # Your analysis here
+           pass
+           
+   except Exception as e:
```
orbeckst, line 259: Catching all exceptions has the potential to make errors go un-noticed.

How about re-raising the exception after closing the trajecorty? Along the lines of
```python
try:
   ...
except Exception as err:
   # log error
   pass
finally:
   u.trajectory.close()

if err:    # won't work, but that's the logic
   raise err
```

doc/source/formats/reference/imd.rst
orbeckst: @ljwoods2 can you raise an issue, please?

doc/source/formats/reference/imd.rst
orbeckst: I would leave out the whole section; perhaps mention somewhere that it may be possible but results are not well defined and link to imdclient docs.

doc/source/formats/reference/imd.rst
orbeckst: use proper section markup for subheadings

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+       # Distance calculations work normally
+       protein = u.select_atoms("protein")
+       rg = protein.radius_of_gyration()
+       
+       # Contact analysis
+       selection1 = u.select_atoms("resid 1-10")
+       selection2 = u.select_atoms("resid 50-60")
+       dist_array = distances.distance_array(selection1.positions, selection2.positions)
+
+**Limitations with Streaming**
```
orbeckst: use proper section markup for subheadings

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+Some analysis methods require the complete trajectory and won't work with streaming:
+
+.. code-block:: python
+
+   # These will NOT work with streaming:
+   # - trajectory.timeseries()
+   # - Most analysis classes that need multiple passes
+   # - Random frame access (trajectory[10])
+   # - Backward iteration
```
orbeckst: Should be a simple list, not a code block

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+Important Limitations
+=====================
+
+Streaming analysis has fundamental constraints due to its real-time nature:
+
+**Data Access Limitations**
```
orbeckst: use proper section markup for subheadings

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+* **No random access**: Cannot jump to arbitrary frames or seek backwards
+* **Forward-only**: Can only iterate through frames as they arrive
+* **Single-use**: Cannot restart iteration once the stream is consumed
+* **No trajectory length**: Total frame count unknown until simulation ends
+* **No independent copies**: Cannot create multiple reader instances for the same stream
+
+**Analysis Constraints**
```
orbeckst: use proper section markup for subheadings

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+
+* **No timeseries methods**: Cannot use ``trajectory.timeseries()``
+* **No bulk operations**: Cannot extract all data at once
+* **Limited multiprocessing**: Cannot split across processes
+* **Single client**: Only one reader per IMD stream (engine-dependent)
+
+**Practical Considerations**
```
orbeckst: use proper section markup for subheadings

doc/source/formats/reference/imd.rst
orbeckst: split into separate code blocks with preceding text that states what is being attempted and what will happen; the comments "will work/won't work" can also stay to make it clear.

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,330 @@
...
+   
+   # This will NOT work - backwards iteration  
+   for ts in u.trajectory[::-1]:  # ValueError
+       pass
+   
+   # This will NOT work - restarting iteration
+   for ts in u.trajectory:
+       break
+   for ts in u.trajectory:  # Won't start from beginning
+       pass
```
orbeckst, line 417: ALso add the `u.trajectory[:10]` example: we cannot (currently) set end-frames.

orbeckst (changes requested):
Great start for the User Guide. I added a bunch of comments in addition to Tyler's and Lawson's ones.

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Made the required change!

doc/source/formats/reference/imd.rst
amruthesht: maybe 100?

doc/source/formats/reference/imd.rst
amruthesht: Shouldn't the MD engine docs be the source here?

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: So tramsit everything in the example, then?

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: @tylerjereddy - nothing specific about IMD requires it. I'll modify the example so it doesn't give the wrong impression

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Added as a comment!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: @ljwoods2 - yes, the keys should be the same for all engines, only implementation availability is engine dependent - I chnage the langauage around it

Thanks for pointing out the erros in the API documentation. I'll make a PR to correct it!

doc/source/formats/reference/imd.rst
amruthesht: @jwoods2 - I actually tested some edge cases out - I'll add it to the relevant PR discussion.

@orbeckst - Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Fixed!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Agreed, thanks for pointing it out!

amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Wouldn't it just start from the next available frame in the buffer?

doc/source/formats/reference/imd.rst
amruthesht: Done!

doc/source/formats/reference/imd.rst
amruthesht: Done - could you check if this is accurate, thanks!

doc/source/formats/reference/imd.rst
ljwoods2: No, it shouldn't if I'm understanding the code right (just trying to run a for loop two times), it should raise an error:

https://github.com/MDAnalysis/mdanalysis/blob/03eef459ddf83e73a776c0e6ddefde4a04ce62c9/testsuite/MDAnalysisTests/coordinates/test_imd.py#L457

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,433 @@
...
+       raise error
+
+Available Data
+--------------
+
+The IMDReader provides access to additional simulation data through ``ts.data``:
+
+* ``dt``: Time step size in picoseconds
+* ``step``: Current simulation step number
+* Energy terms: ``potential_energy``, ``total_energy``, etc. (IMD-streamed in NAMD and GROMACS only)
```
ljwoods2: Nitpick, but maybe say available as an option in NAMD and GROMACS only
```suggestion
* Energy terms: ``potential_energy``, ``total_energy``, etc. (available as an option in NAMD and GROMACS only)
```

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,433 @@
...
+  - Between-frames: :func:`~MDAnalysis.lib.distances.calc_bonds` for displacement calculations comparing consecutive frames
+  - Between-frames: Frame-to-frame RMSD calculations using :func:`~MDAnalysis.analysis.rms.rmsd` with stored reference coordinates
+  - Within-frame: Real-time monitoring using :meth:`~MDAnalysis.core.groups.AtomGroup.center_of_mass` for quality control checks
+
+* **Accumulative analyses**: Building results incrementally across frames using :class:`~MDAnalysis.analysis.base.AnalysisBase` patterns, for example:
+
+  - :class:`~MDAnalysis.analysis.rdf.InterRDF` - Frame-by-frame radial distribution function calculations
+  - :class:`~MDAnalysis.analysis.dihedrals.Dihedral` - Dihedral angle accumulation for conformational analysis  
+  - :class:`~MDAnalysis.analysis.lineardensity.LinearDensity` - Density profile building over streaming frames
+
```
ljwoods2: Thanks for writing this up, adds a ton of clarity for users

Read through the analysis classes, agree these are good examples of what will work.

One small suggestion, should specify that backend has to be serial
```suggestion
.. note::
    Passing any backend other than ``backend="serial"`` to these analysis classes will cause them to fail since streams are single-pass and forward-only.
```

ljwoods2 (changes requested):

doc/source/formats/reference/imd.rst
```diff
@@ -0,0 +1,433 @@
...
+       
+       # Your analysis code here
+       center_of_mass = selected_atoms.center_of_mass()
+       print(f"Protein COM: {center_of_mass}")
+       
+       # Optional: break on some condition
+       if ts.time > 1000:  # Stop after 1000 ps
+           break
+
+The ``buffer_size`` parameter (10 MB = 10*1024*1024 bytes in this example) controls the buffer used by `imdclient <https://imdclient.readthedocs.io/>`_ to temporarily store data received from the socket. This buffer accounts for speed differences between the producer (simulation engine) and receiver (analysis code), preventing data loss when analysis is slower than data transmission. A larger buffer is more suitable for systems with many atoms or high transmission frequencies. For more details on buffer management and optimization, see the `imdclient documentation <https://imdclient.readthedocs.io/>`_.
```
orbeckst, line 186: Generally I would try to replace in-line hyperlinks with named links just to keep things tidy, especially if you're using the same URL multiple times. (Optional)

doc/source/formats/reference/imd.rst
orbeckst: Would still recommend breaking up this example


> I'd remove the optional part from this example so that it can stand as a simple example on its own.
> 
> Then create another example where you discuss the issue with how breaking the loop may affect the running simulation. This is important but nuanced. If something is complicated, break it into smaller steps, that's basic pedagogy.

orbeckst (changes requested):
The reST document looks pretty good to me (I think one major comment is unresolved). The notebook still has many unresolved comments as far as I can tell.

orbeckst:
RTD installation fails because 3.10 not supported by MDA anymore. (Raised #431 )

orbeckst:
@amruthesht a simple thing to address are the issues from pre-commit https://github.com/MDAnalysis/UserGuide/actions/runs/18957769421/job/54138651607?pr=430

tylerjereddy:
I'm still a bit confused by the indication:

>  MDAnalysis supports **IMDv3 only**, which uses the IMDv3 protocol to provide continuous, gap-free streaming and is implemented in modern versions of GROMACS

but I just tried it with GMX `2025.3` (I actually have a genuine use case), and I see:

> Unknown left-hand 'IMD-version' in parameter file

and then I see: https://imdclient.readthedocs.io/en/stable/usage.html#gromacs

> The IMDv3 protocol is currently not available as part of the official GROMACS release or source code repository. However, the feature is currently available for use in the imd-v3 branch of the forked repository

That's a pretty substantial distinction, both in convenience (I often pull in GMX binaries from `conda-forge`) and trust (reviewed/merged by GMX core devs). It looks like this issue is relevant: https://gitlab.com/gromacs/gromacs/-/issues/5311

orbeckst:
>> MDAnalysis supports IMDv3 only, which uses the IMDv3 protocol to provide continuous, gap-free streaming and is implemented in modern versions of GROMACS

This needs to be changed to accurately reflect current availability. I would just say something along the lines of 

> For current implementation status of IMDv3 in major MD codes see [Configuring the Simulation Engine](https://imdclient.readthedocs.io/en/stable/usage.html#configuring-the-simulation-engine-for-imdv3) in the imdclient documentation.

In this way we only need to keep one piece of text up-to-date.

doc/source/examples/other/streaming_imd.ipynb
orbeckst: Note: imdclient is an _optional_ dependency so it's not automatically installed.

orbeckst:
@tylerjereddy based on your comments, @amruthesht is working on IMDv2 support in imdclient (PR Becksteinlab/imdclient#141 ) and once that works, we can update the MDA IMDReader to also support it. It will not be a full-feature experience (for instance, box information is not available) but it may be sufficient for "playing around" or simulation monitoring with [mdadash](https://github.com/MDAnalysis/mdadash).

orbeckst:
@amruthesht with the paper accepted, can you finish up the UG entry?

